PR #4, opened 2020-07-02
FAT2-79 Change maxfunding mapping

dashton82:
Added filter so that only standards that have a correct start date are displayed. Change the MaxFunding amount to come from the collection on the response.


src/SFA.DAS.FindApprenticeshipTraining.UnitTests/Application/TrainingCourses/Queries/WhenGettingTrainingCourseList.cs
```diff
@@ -39,6 +43,240 @@ public class WhenGettingTrainingCourseList
...
+            mockApiClient
+                .Setup(client => client.Get<GetSectorsListResponse>(It.IsAny<GetSectorsListRequest>()))
+                .ReturnsAsync(sectorsApiResponse);
+            
+            //Act
+            var result = await handler.Handle(query, CancellationToken.None);
+            
+            //Assert
+            Assert.IsTrue(result.Courses.ToList().TrueForAll(c=>c.MaxFunding.Equals(fundingPrice)));
+        }
```
scottywakefield, line 278: these tests are more to do with the GetStandardsListItem - put them on that and the tests will be much simpler.

dashton82: Good idea - I've moved those two tests over that are specific to the `GetStandardsListItem`

PR #9, opened 2020-07-08
Ei 75 create employer incentives outer api

Pauljgraham:


src/SFA.DAS.EmployerIncentives/Infrastructure/Api/RestApiClient.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Net.Http;
+using System.Net.Http.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
+using SFA.DAS.EmployerIncentives.Interfaces;
+
+namespace SFA.DAS.EmployerIncentives.Infrastructure.Api
+{
+    public class RestApiClient : IRestApiClient
```
David-B-Read, line 12: There's a RestHttpClient in the SFA.DAS.Http Nuget package, can that be inherited from instead?

src/SFA.DAS.Apim.Endpoints.v3.ncrunchsolution
```diff
@@ -0,0 +1,6 @@
+﻿<SolutionConfiguration>
```
dashton82, line 1: This file can be ignored

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using NUnit.Framework;
+
+namespace SFA.DAS.EmployerIncentives.Api.AcceptanceTests.Pipeline
+{
+    [TestFixture]
+    [Parallelizable]
+    public class WhenCheckingPipeLine
+    {
+        [Test]
+        [Ignore("being reworked")]
```
dashton82, line 15: Why is this ignored? The test should have been updated to accommodate the rework

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        public async Task CallPingEndpoint_ThenShouldReturnOkResponse()
+        {
+            var f = new HealthCheckFixture();
+            var r =  await f.Client.GetAsync("/ping");
+            r.StatusCode.Should().Be(HttpStatusCode.OK);
+        }
+
+        [Test]
+        [Ignore("being reworked")]
```
dashton82, line 24: Why is this ignored? The test should have been updated to accommodate the rework

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        public async Task CallHealthEndpoint_ThenShouldReturnOkResponse()
+        {
+            var f = new HealthCheckFixture();
+            var r = await f.Client.GetAsync("/health");
+            r.StatusCode.Should().Be(HttpStatusCode.OK);
+        }
+    }
+
+    public class HealthCheckFixture
```
dashton82, line 33: No Test fixture files - These become massively bloated and confusing. Please use the SFA.DAS.Testing package for any autofixture creating and mocking.

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.6.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.1.5" />
+    <!--<PackageReference Include="Microsoft.AspNetCore.All" Version="2.2.1" />
```
dashton82, line 10: Remove

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />-->
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="Moq" Version="4.12.0" />
+    <PackageReference Include="NUnit" Version="3.11.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.12.0" />
+    <!--<PackageReference Include="WireMock.Net" Version="1.2.12" />-->
```
dashton82, line 16: Remove

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.6.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.1.5" />
+    <!--<PackageReference Include="Microsoft.AspNetCore.All" Version="2.2.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />-->
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
```
dashton82, line 12: These packages all need updating to latest

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using SFA.DAS.EmployerIncentives.Models.PassThrough;
+
+namespace SFA.DAS.EmployerIncentives.Api.UnitTests.Controllers
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    public class PassThroughControllerTests
+    {
+        [Test]
+        public async Task When_AddLegalEntity_is_requested_Then_EmployerIncentivesPassThroughService_is_called()
```
dashton82, line 19: All Test names have inconsistent capitalisation.

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            var f = new TestsFixture().SetupEmployerIncentivesServiceForAddLegalEntityCall();
```
dashton82, line 21: No fixtures - these are bloated, complex and confusing. Use the SFA.DAS.Testing package for any auto mocking

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                Sut = new PassThroughController(EmployerIncentivesPassThroughServiceMock.Object);
+
+                AccountId = Fixture.Create<long>();
+                AccountLegalEntityId = Fixture.Create<long>();
+                LegalEntityRequest = Fixture.Create<LegalEntityRequest>();
+                InnerApiResponse = Fixture.Build<InnerApiResponse>().Without(x=>x.Json).Create();
+            }
+
+            public TestsFixture SetupEmployerIncentivesServiceForAddLegalEntityCall()
```
dashton82, line 81: None of these seem very specific to the actual request, we shouldn't be doing It.IsAny for anything other than checking that something is never called

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
MichaelYoung1981, line 10: Can the commented out references be deleted?

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        [Test]
+        public async Task When_AddLegalEntity_is_requested_Then_EmployerIncentivesPassThroughService_is_called()
+        {
+            var f = new TestsFixture().SetupEmployerIncentivesServiceForAddLegalEntityCall();
+
+            await f.Sut.AddLegalEntity(f.AccountId, f.LegalEntityRequest);
+
+            f.EmployerIncentivesPassThroughServiceMock.Verify(x => x.AddLegalEntity(f.AccountId, f.LegalEntityRequest));
+        }
```
MichaelYoung1981, line 26: Bit of an interesting one this...we probably need to make sure the same conventions and tool are used across the solution even though that may differ from what we have agreed on EI.

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
MichaelYoung1981, line 19: I'd rename the tests so they don't describe the implementartion.

src/SFA.DAS.EmployerIncentives.Api/AppStart/ServiceRegistrationExtension.cs
```diff
@@ -0,0 +1,44 @@
...
+            services.AddTransient<IRestApiClient, RestApiClient>();
+
+            var httpBuilder = services.AddHttpClient("EmployerIncentivesInnerApi", c =>
+                {
+                    var apiConfig = GetConfigSection(configuration, EmployerIncentivesConfigurationKeys.EmployerIncentivesInnerApiConfiguration);
+                    c.BaseAddress = new Uri(apiConfig.Url);
+                })
+                .AddTypedClient<IEmployerIncentivesPassThroughService, EmployerIncentivesPassThroughService>();
+
+            if (!env.IsDevelopment())
```
MichaelYoung1981, line 27: This method can be a PITA!  If you want to run the outer locally but with the inner from AT it won't work.  I've always found IsLoopback on the API base address to be easier.

src/SFA.DAS.EmployerIncentives.Api/appsettings.acceptancetests.json
```diff
@@ -0,0 +1,9 @@
+{
+  "SFA.DAS.EmployerIncentives.OuterApi": {
```
dashton82, line 2: Shouldnt need this file - any configuration you require for acceptance tests should be done using the InMemoryCollection

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using SFA.DAS.EmployerIncentives.Configuration;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authentication
+{
+    public static class AuthenticationExtensions
```
MichaelYoung1981, line 10: Does this class, or similar, not exist in one of the projects already in the solution?  If it does, can we reuse that instead?

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            services.AddControllers(c=>c.Filters.Add(new AuthorizeFilter("APIM")))
+                .AddJsonOptions(options => {
+                    options.JsonSerializerOptions.PropertyNameCaseInsensitive = true;
+                }); 
+
+            services.AddDasHttpClientsAndAssociatedServices(_configuration, _env);
+        }
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
```
dashton82, line 67: Should be IWebHostEnvironment

src/SFA.DAS.EmployerIncentives.Api/Authentication/AzureAdScopeClaimTransformation.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authentication
+{
+    public class AzureAdScopeClaimTransformation : IClaimsTransformation
```
MichaelYoung1981, line 10: As per previous comment, looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authorization
+{
+    public static class ApiAuthorizationExtensions
```
MichaelYoung1981, line 9: As per previous, looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationPolicyBuilderExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.AspNetCore.Authorization;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authorization
+{
+    public static class AuthorizationPolicyBuilderExtensions
```
MichaelYoung1981, line 5: As per previous, looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Authorization/LocalAuthorizationHandler.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authorization
+{
+    public class LocalAuthorizationHandler : AuthorizationHandler<NoneRequirement>
+    {
+        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context,
```
MichaelYoung1981, line 8: As per previous, looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                config.AddJsonFile($"appsettings.development.json", optional: true);
+            }
+
+            if (_env.IsEnvironment("AcceptanceTests"))
+            {
+                config.AddJsonFile($"appsettings.acceptancetests.json", optional: false);
+            }
+
+            _configuration = config.Build();
```
dashton82, line 41: You've not add the `.AddEnvironmentVariables();` so its unlikely this will work once deployed

src/SFA.DAS.EmployerIncentives.Api/Authorization/NoneRequirement.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Microsoft.AspNetCore.Authorization;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authorization
+{
+    public class NoneRequirement : IAuthorizationRequirement
```
MichaelYoung1981, line 5: As per previous, looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Controllers/PassThroughController.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.EmployerIncentives.Interfaces;
+using SFA.DAS.EmployerIncentives.Models.PassThrough;
+
+namespace SFA.DAS.EmployerIncentives.Api.Controllers
+{
+    [ApiController]
+    public class PassThroughController : ControllerBase
```
MichaelYoung1981, line 9: I'm not convinced on the name as this will get massive!  I'd rather preserve the same controller structure as we have in the inner API.

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+                .AddConfiguration(configuration);
+
+            if (_env.IsDevelopment())
+            {
+                config.AddJsonFile($"appsettings.development.json", optional: true);
+            }
+
+            if (_env.IsEnvironment("AcceptanceTests"))
+            {
+                config.AddJsonFile($"appsettings.acceptancetests.json", optional: false);
```
dashton82, line 38: This shouldn't be here

src/SFA.DAS.EmployerIncentives.Api/ErrorHandler/ErrorHandlerExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Diagnostics;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+
+namespace SFA.DAS.EmployerIncentives.Api.ErrorHandler
+{
+    public static class ErrorHandlerExtensions
+    {
+        public static IApplicationBuilder UseApiGlobalExceptionHandler(this IApplicationBuilder app, ILogger logger)
```
MichaelYoung1981, line 12: Looks like this could be shared?

src/SFA.DAS.EmployerIncentives.Api/Extensions/HttpResponseExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace SFA.DAS.EmployerIncentives.Api.Extensions
+{
+    public static class HttpResponseExtensions
+    {
+        public static Task WriteJsonAsync(this HttpResponse httpResponse, object body)
```
MichaelYoung1981, line 11: Looks like this could be shared.

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.OpenApi.Models;
+using SFA.DAS.EmployerIncentives.Api.AppStart;
+using SFA.DAS.EmployerIncentives.Api.Authentication;
+using SFA.DAS.EmployerIncentives.Api.Authorization;
+using SFA.DAS.EmployerIncentives.Api.Configuration;
+using SFA.DAS.EmployerIncentives.Api.ErrorHandler;
+using SFA.DAS.EmployerIncentives.Api.HealthChecks;
+using IHostingEnvironment = Microsoft.AspNetCore.Hosting.IHostingEnvironment;
```
dashton82, line 15: This shouldnt be here

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (_env.IsDevelopment())
+            {
+                config.AddJsonFile($"appsettings.development.json", optional: true);
+            }
+
+            if (_env.IsEnvironment("AcceptanceTests"))
+            {
+                config.AddJsonFile($"appsettings.acceptancetests.json", optional: false);
+            }
+
```
MichaelYoung1981, line 40: Having code specific for tests makes me sad on the inside.

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace SFA.DAS.EmployerIncentives.Api
+{
+    public class Startup
+    {
+        private readonly IWebHostEnvironment _env;
+        private readonly IConfiguration _configuration;
+
+        public Startup(IConfiguration configuration, IWebHostEnvironment env)
+        { _env = env;
+            _configuration = configuration;
```
dashton82, line 26: This should be added to the ConfigurationBuilder

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,98 @@
...
+            _configuration = config.Build();
+        }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddApiConfigurationSections(_configuration);
+            services.AddApiAuthentication(_configuration);
+            services.AddApiAuthorization(_env);
+
+            // Should this be upgraded?
```
MichaelYoung1981, line 50: Probably.

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                .AddCheck<CommitmentsV2ApiHealthCheck>("Commitments V2 API Health Check");
+
+            return services;
+        }
+
+        public static IApplicationBuilder UseDasHealthChecks(this IApplicationBuilder app)
+        {
+            app.UseHealthChecks("/ping"); 
+
+            return app.UseHealthChecks("/health", new HealthCheckOptions
```
dashton82, line 24: There already is a HealthCheckResponseWriter that should be used

src/SFA.DAS.EmployerIncentives/Services/EmployerIncentivesPassThroughService.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using SFA.DAS.EmployerIncentives.Infrastructure.Api;
+using SFA.DAS.EmployerIncentives.Interfaces;
+using SFA.DAS.EmployerIncentives.Models.PassThrough;
+
+namespace SFA.DAS.EmployerIncentives.Services
+{
+    public class EmployerIncentivesPassThroughService : IEmployerIncentivesPassThroughService
```
MichaelYoung1981, line 10: This should be specific to the inner api endpoint it's calling otherwise it'll become massive.

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Net.Http.Json;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
+using SFA.DAS.EmployerIncentives.Interfaces;
+
+namespace SFA.DAS.EmployerIncentives.Infrastructure.Api
+{
+    public class PassThroughApiClient : IPassThroughApiClient
```
MichaelYoung1981, line 13: Does this class give us anything above the RestApiClient?  The methods on here look like they could sit on there instead.

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/InnerApiHealthCheck.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace SFA.DAS.EmployerIncentives.Api.HealthChecks
+{
+    public class InnerApiHealthCheck : IHealthCheck
```
dashton82, line 7: Doesn't seem very useful this healthcheck

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/ManagedIdentityApiHandler.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Services.AppAuthentication;
+using SFA.DAS.EmployerIncentives.Configuration;
+
+namespace SFA.DAS.EmployerIncentives.Infrastructure.Api
+{
+    public class ManagedIdentityApiHandler : DelegatingHandler
```
MichaelYoung1981, line 10: This looks like it could be shared.

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/CommitmentsV2ApiHealthCheck.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace SFA.DAS.EmployerIncentives.Api.HealthChecks
+{
+    public class CommitmentsV2ApiHealthCheck : IHealthCheck
+    {
+        public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = new CancellationToken())
```
dashton82, line 9: Doesn't seem very useful this healthcheck

src/SFA.DAS.EmployerIncentives.UnitTests/Services/EmployerIncentivesPassThroughServiceTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            var f = new TestsFixture().WithNoJsonResponseFromInnerApi();
+            var result = await f.Sut.RemoveLegalEntity(f.AccountId, f.AccountLegalEntityId);
+
+            result.Should().NotBeNull();
+            result.StatusCode.Should().Be(f.HttpResponseMessage.StatusCode);
+            result.Json.Should().BeNull();
+        }
+
+        private class TestsFixture
```
MichaelYoung1981, line 59: Probably worth having a discussion with Dan and other EI devs to see if we use this approach?  As far as I'm aware, we aren't using it in any of the other EI solutions?

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+                    else
+                    {
+                        policy.RequireAuthenticatedUser();
+                        policy.AddAuthenticationSchemes(JwtBearerDefaults.AuthenticationScheme);
+                        policy.RequireRole("APIM");
+                    }
+                });
+            });
+            if (isDevelopment)
+                services.AddSingleton<IAuthorizationHandler, LocalAuthorizationHandler>();
```
dashton82, line 32: Why is this required? Why not just not have Auth if you are local?

src/SFA.DAS.EmployerIncentives.Api/Authentication/AzureAdScopeClaimTransformation.cs
dashton82, line 10: This is already defined in the solution

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public static IServiceCollection AddApiAuthentication(this IServiceCollection services, IConfiguration config)
```
dashton82, line 12: This is already defined elsewhere in the solution

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/RestApiClient.cs
Pauljgraham: We are not using the Http Package as it's got references to Structure map and the httpClient. Also because I didn't think it had been reviewed. I added the updates to this branch to make core 3.1. Would you kindly take another look thx

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: yes

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
Pauljgraham: Ok will discuss with the Dan

src/SFA.DAS.EmployerIncentives.Api/AppStart/ServiceRegistrationExtension.cs
Pauljgraham: Don't think IsLoopback works in core. I used in framework, but found I couldn't implement it in Core.   But point taken about mixing environments

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
Pauljgraham: Thjere was no shared classes in the code when I started. Will discuss with Dan, See what they are doing

src/SFA.DAS.EmployerIncentives.Api/Authentication/AzureAdScopeClaimTransformation.cs
Pauljgraham: Agreed

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationExtensions.cs
Pauljgraham: agree

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationPolicyBuilderExtensions.cs
Pauljgraham: agreed

src/SFA.DAS.EmployerIncentives.Api/Authorization/LocalAuthorizationHandler.cs
Pauljgraham: yes

src/SFA.DAS.EmployerIncentives.Api/Authorization/NoneRequirement.cs
Pauljgraham: yes

src/SFA.DAS.EmployerIncentives.Api/ErrorHandler/ErrorHandlerExtensions.cs
Pauljgraham: agree

src/SFA.DAS.EmployerIncentives.Api/Controllers/PassThroughController.cs
Pauljgraham: The intent of this controller is to simply pass forward the request to the inner api, where we do not need to do anything to it.
It would be worth splitting this into queries requests and non-query requests. As polly can be applied to the request queries.

The name is worthwhile discussing but the intent of the code here is to quickly forward it to the inner API and return the response.

src/SFA.DAS.EmployerIncentives.Api/Extensions/HttpResponseExtensions.cs
Pauljgraham: yes

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: I know. But for the FAKE stubs to work we need to override teh default config setting in AzureStorage. I guess the file will get loaded in an acceptancetest environment, but we need to place it here so it is loaded last.

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: I'm looking into this now

src/SFA.DAS.EmployerIncentives/Services/EmployerIncentivesPassThroughService.cs
Pauljgraham: It is the intention of this class is to simply forward inner api requests on. Perhaps we should call it

EmployerIncentivesInnerApiPassThroughService

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
Pauljgraham: Yes, it streamlines the json coming back from the inner api and just produces a json response. If we try to do this with the RestApiClient. We would have to add plumbing work simply to mirror the return types for each API call.

Because we are adding no value to the passthrough requests, it seems wrong to do anything but forward the response.

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/ManagedIdentityApiHandler.cs
Pauljgraham: yes

src/SFA.DAS.EmployerIncentives.UnitTests/Services/EmployerIncentivesPassThroughServiceTests.cs
Pauljgraham: Will do

src/SFA.DAS.Apim.Endpoints.v3.ncrunchsolution
Pauljgraham: ok, will ignore it in git files

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
Pauljgraham: oh because we are planning to rewrite the Acceptance tests in SpecFlow and these currently are in NUnit

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
Pauljgraham: as above

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
Pauljgraham: Can we discuss on monday?

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: will do

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: will do

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: ok

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
Pauljgraham: Happy to follow standards in the project.

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
Pauljgraham: The only thing I would say, is they can keep the tests down to a few lines and it 'sometimes' reads much better

src/SFA.DAS.EmployerIncentives.Api.UnitTests/Controllers/PassThroughControllerTests.cs
Pauljgraham: Perhaps these can be tested fully via the Acceptance Tests

src/SFA.DAS.EmployerIncentives.Api/appsettings.acceptancetests.json
Pauljgraham: It's to override the AzureStorage Configs. Our acceptance tests are not going to call the External APIs.   We are going to use wiremock to test the API calls, but we need to override the config values

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: done

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: good point

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: Can we discuss

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: Agreed

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/ServiceCollectionExtensions.cs
Pauljgraham: Can we discuss?

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/InnerApiHealthCheck.cs
Pauljgraham: This was a stub, it's now got an implementation

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/CommitmentsV2ApiHealthCheck.cs
Pauljgraham: same as above

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationExtensions.cs
Pauljgraham: This allow the Authentication pipe line to be called for an anonymous user in the same way as a authenticated user. In theory it will allow us to add the [Authorized] to controllers and it would still work. However, can we discuss the best way forward

src/SFA.DAS.EmployerIncentives.Api/Authentication/AzureAdScopeClaimTransformation.cs
Pauljgraham: I didn't see any shared projects?

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
Pauljgraham: Can we discuss?

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/Pipeline/HealthCheckTests.cs
Pauljgraham: removed

src/SFA.DAS.EmployerIncentives.Api/AppStart/ServiceRegistrationExtension.cs
Pauljgraham: These extension methods will be consolidated into the shared project in the near future

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
Pauljgraham: Will be in the shared project eventually

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
Pauljgraham: shared project

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/CommitmentsV2ApiHealthCheck.cs
Pauljgraham: fixed

src/SFA.DAS.EmployerIncentives.Api/HealthChecks/InnerApiHealthCheck.cs
Pauljgraham: fixed

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: Will look at how to remove this into memory variables

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: resolved

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: Will look at moving these into memory variables for acceptance tests

src/SFA.DAS.EmployerIncentives/Interfaces/IRestApiClient.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerIncentives.Interfaces
+{
+    public interface IRestApiClient
+    {
+        Task<string> GetAsync(Uri uri, object queryData = null, CancellationToken cancellationToken = default);
```
David-B-Read, line 9: I think the convention here is to only use the Async suffix when there's a non-async version of the method as well?

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
```diff
@@ -0,0 +1,94 @@
...
+            return await CreateApiResponse(response);
+        }
+
+        public async Task<InnerApiResponse> DeleteAsync(string uri, CancellationToken cancellationToken = default)
+        {
+            var response = await _httpClient.DeleteAsync(uri, cancellationToken).ConfigureAwait(false);
+            return await CreateApiResponse(response);
+        }
+
+        private async Task<InnerApiResponse> CreateApiResponse(HttpResponseMessage responseMessage)
```
Stephek, line 58: Do we need this?  Is there a reason not to just return the HttpResponse?

src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
Pauljgraham: will be done in Accept Tests branch

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
Pauljgraham: done

src/SFA.DAS.EmployerIncentives/Interfaces/IRestApiClient.cs
Pauljgraham: fixing

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        public async Task<InnerApiResponse> Get(Uri uri, object queryData = null, CancellationToken cancellationToken = default)
+        {
+            if (queryData != null)
+            {
+                uri = new Uri(AddQueryString(uri.ToString(), queryData), UriKind.RelativeOrAbsolute);
+            }
+
+            var response = await _httpClient.GetAsync(uri, cancellationToken).ConfigureAwait(false);
```
Stephek, line 32: You don't need ConfigureAwait in .net core 3.1 since there is no Synchronization context.

src/SFA.DAS.EmployerIncentives.Api/Extensions/HttpResponseExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            httpResponse.ContentType = "application/json";
+
+            return httpResponse.WriteAsync(JsonConvert.SerializeObject(body, new JsonSerializerSettings
```
Stephek, line 18: Avoid using Newtonsoft.Json to load whole document into memory.  Use System.Text.Json that is part of core 3.1 instead to stream the data.

src/SFA.DAS.EmployerIncentives.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,45 @@
...
+        public AccountController(IPassThroughApiClient client)
+        {
+            _client = client;
+        }
+
+        [HttpGet]
+        [Route("/accounts/{accountId}/legalentities")]
+        public async Task<IActionResult> GetLegalEntities(long accountId)
+        {
+            var innerApiResponse = await _client.Get($"/accounts/{accountId}/legalentities");
```
Stephek, line 23: Is the purpose of CreateObjectResult to just set the header?  If you use HttpCompletionOption.ResponseHeadersRead as an option then the headers can be read without loading the body.  Then the body should be able to be streamed to the client without loading in memory.

src/SFA.DAS.EmployerIncentives.Api/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using SFA.DAS.EmployerIncentives.Configuration;
+
+namespace SFA.DAS.EmployerIncentives.Api.Configuration
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddApiConfigurationSections(this IServiceCollection services, IConfiguration configuration)
+        {
+            services.AddOptions();
+            services.Configure<AzureActiveDirectoryApiConfiguration>(configuration.GetSection(EmployerIncentivesConfigurationKeys.AzureActiveDirectoryApiConfiguration));
```
dashton82, line 12: This configuration class is in shared and can be reused here

src/SFA.DAS.EmployerIncentives.Api/Startup.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace SFA.DAS.EmployerIncentives.Api
+{
+    public class Startup
+    {
+        private readonly IWebHostEnvironment _env;
+        private readonly IConfiguration _configuration;
+
+        public Startup(IConfiguration configuration, IWebHostEnvironment env)
+        { _env = env;
+            _configuration = configuration;
```
dashton82, line 25: Probably don't need to se this here? Its set further down from the config.Build()

src/SFA.DAS.EmployerIncentives/Configuration/AzureActiveDirectoryApiConfiguration.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SFA.DAS.EmployerIncentives.Configuration
+{
+    public class AzureActiveDirectoryApiConfiguration
```
dashton82, line 3: You can reuse the one in the shared project

src/SFA.DAS.EmployerIncentives/Configuration/AzureManagedIdentityApiConfiguration.cs
```diff
@@ -0,0 +1,11 @@
+﻿using SFA.DAS.EmployerIncentives.Interfaces;
+
+namespace SFA.DAS.EmployerIncentives.Configuration
+{
+    public class AzureManagedIdentityApiConfiguration
```
dashton82, line 5: This should be the same as the above - the Url isn't required - its only the method of obtaining the token that is different

src/SFA.DAS.EmployerIncentives.Api/Authentication/AuthenticationExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            var azureActiveDirectoryConfiguration = config.GetSection(EmployerIncentivesConfigurationKeys.AzureActiveDirectoryApiConfiguration).Get<AzureActiveDirectoryApiConfiguration>();
```
dashton82, line 14:
```suggestion
            var azureActiveDirectoryConfiguration = config.GetSection("AzureAd").Get<AzureActiveDirectoryApiConfiguration>();
```
You don't need to pass the whole path, to avoid the ConfigurationKeys constants you could change your config name and use `nameof(AzureActiveDirectoryApiConfiguration)'

I'm easy either way 😄

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
dashton82: I agree - I think I'm also struggling with the concept of this class - are you using everything that comes back from the response in the front end? 
For `POST` and `DELETE` I'm unsure as to why you can't use the other client?

src/SFA.DAS.EmployerIncentives/Services/CommitmentsV2Service.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace SFA.DAS.EmployerIncentives.Services
+{
+    public class CommitmentsV2Service : ICommitmentsV2Service
+    {
+        private readonly IRestApiClient _restApiClient;
+
+        public CommitmentsV2Service(HttpClient client)
+        {
+            _restApiClient = new RestApiClient(client);
```
dashton82, line 16: This doesn't seem right to me 👀

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/RestApiClient.cs
```diff
@@ -0,0 +1,87 @@
...
```
dashton82, line 12: Is there any reason the differences in here can't be added to the ApiClient in shared? from what i can tell the only difference is the Cancellation token which is just being set to default - not changed - and is the default for if you don't pass it.

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
Pauljgraham: The intent of the class is to keep it as light as possible. I worry about returning the HttpResponse from this class as it's content is a stream and I'm unsure what happens to it if when the client goes out of scope.

I don't want to use the other client as this means we have to start putting more infrastructure class in place to deal with the different types of responses. Such as validation messages

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/PassThroughApiClient.cs
Pauljgraham: Agreed. but in the long term I believe this probably will become a nuget standard package and will therefore need it for framework code

src/SFA.DAS.EmployerIncentives.Api/Extensions/HttpResponseExtensions.cs
Pauljgraham: agreed. Will remove

src/SFA.DAS.EmployerIncentives.Api/Authorization/AuthorizationExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace SFA.DAS.EmployerIncentives.Api.Authorization
```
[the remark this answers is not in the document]
Pauljgraham: Will do

src/SFA.DAS.EmployerIncentives/Services/CommitmentsV2Service.cs
Pauljgraham: This is due to the way we need to wire up the clients. If you look at how we add the HttpClients in AppStart\ServiceRegistrationExtensions.cs

You can see how we are adding clients.

src/SFA.DAS.EmployerIncentives/Infrastructure/Api/RestApiClient.cs
Pauljgraham: No it's quite different from ApiClient. We are adding the authentication and logging as HttpMessageHandlers and this seems to make the intent of the each layer clearer.

PR #24, opened 2020-07-20
Fat2 134 sorting

PeterDavidWilliams:


src/SFA.DAS.FindApprenticeshipTraining.Api/Controllers/TrainingCoursesController.cs
```diff
@@ -27,15 +28,16 @@ public class TrainingCoursesController : ControllerBase
         }
 
         [HttpGet]
-        public async Task<IActionResult> GetList( [FromQuery] string keyword = "", [FromQuery] List<Guid> routeIds = null, [FromQuery]List<int> levels = null)
+        public async Task<IActionResult> GetList( [FromQuery] string keyword = "", [FromQuery] List<Guid> routeIds = null, [FromQuery]List<int> levels = null, [FromQuery]int orderBy = 0)
```
scottywakefield, line 31: web is sending this as a string. semantic web and all.

src/SFA.DAS.FindApprenticeshipTraining.Api/Controllers/TrainingCoursesController.cs
```diff
@@ -27,15 +28,16 @@ public class TrainingCoursesController : ControllerBase
...
         {
             try
             {
                 var queryResult = await _mediator.Send(new GetTrainingCoursesListQuery
                 {
                     Keyword = keyword, 
                     RouteIds = routeIds,
-                    Levels = levels
+                    Levels = levels,
+                    OrderBy = (OrderBy)orderBy
```
scottywakefield, line 40: having a string from front end can be mapped to the web's enum, and then map to inner api enum. don't assume that they will map 0==0, 1==1. be explicit. else there'll be mysterious bugs one day when a new order by is added or removed.

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetStandardsListRequest.cs
```diff
@@ -15,6 +14,7 @@ public class GetStandardsListRequest : IGetApiRequest
         public string GetUrl => BuildUrl();
 
         public List<int> Levels { get ; set ; }
+        public OrderBy OrderBy { get; set; }
```
scottywakefield, line 17: won't you need that in build url as well?

src/SFA.DAS.FindApprenticeshipTraining.Api/Controllers/TrainingCoursesController.cs
dashton82: This is fixed

src/SFA.DAS.FindApprenticeshipTraining.Api/Controllers/TrainingCoursesController.cs
dashton82: This is also fixed

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetStandardsListRequest.cs
dashton82: Updated to be in the request to the inner api

PR #32, opened 2020-07-27
Ei 167 add jobs endpoint branched from ei 85

Stephek:


src/SFA.DAS.EmployerIncentives.Api.AcceptanceTests/SFA.DAS.EmployerIncentives.Api.AcceptanceTests.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <SpecFlowObsoleteCodeBehindFiles Remove="Features\LegalEntityRequests - Copy.feature.cs" />
```
Pauljgraham, line 8: I don't think this should be here

Stephek: Good spot.  Will remove.

Stephek:
> One tiny issue with a project file. See comment
> 
> Also I'm intrigued why we expose this functionality as an endpoint. Doing so allows the calling applications to repopulate the AccountLegalEntities. I haven#t looked at the inner API functionality, so this may not be a issue. I just imagined the job would simply run once when the code was first deployed and would be never needed again

Implemented as per the requirement.

PR #39, opened 2020-07-31
FAT2-112 provider details

dashton82:


dashton82:
Not needed - already merged in

PR #46, opened 2020-08-03
Fat2 118 health checks

PeterDavidWilliams:


src/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPingRequest.cs
```diff
@@ -0,0 +1,10 @@
+using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.EmployerIncentives.InnerApi.Requests
```
scottywakefield:
```suggestion
namespace SFA.DAS.SharedOuterApi.InnerApi.Requests
```
this might break some references so best do it using resharper

PR #74, opened 2020-08-11
EI-145 add endpoint to send bank details required email

David-B-Read:


src/SFA.DAS.EmployerIncentives/Application/Services/EmployerIncentivesService.cs
```diff
@@ -96,6 +104,13 @@ public async Task<IncentiveApplicationDto> GetApplication(long accountId, Guid a
             return response;
         }
 
+        public async Task<long> GetApplicationLegalEntity(long accountId, Guid applicationId)
+        {
+            var response = await _client.Get<long>(new GetApplicationLegalEntityRequest(accountId, applicationId));
+
+            return response;
+        }
+
```
Pauljgraham, line 113: would it be better to simply get the LegalEntityId from the Application. As it's being added there. Saves this extra call

src/SFA.DAS.EmployerIncentives.Api/Controllers/ApplicationController.cs
```diff
@@ -64,5 +65,18 @@ public async Task<IActionResult> GetApplication(long accountId, Guid application
...
+        public async Task<IActionResult> GetApplicationLegalEntity(long accountId, Guid applicationId)
+        {
+            var result = await _mediator.Send(new GetApplicationLegalEntityQuery
+            {
+                AccountId = accountId,
+                ApplicationId = applicationId
+            });
+
+            return Ok(result);
+        }
```
Pauljgraham, line 80: A couple of concerns here result appears to be an int which we pass back (not inside an object). But my main concern is since Mike has added the LegalEntityId to the GetApplication call, we could simply call the 'GetApplicationQuery' and pick up the value from there. However if we extend the GetApplication controller result to include the LegalEntityId then this endpoint is not needed at all.

src/SFA.DAS.EmployerIncentives/Application/Services/EmployerIncentivesService.cs
David-B-Read: Two reasons for this endpoint: One is that it seems overkill to pass large amounts of JSON (full application and all the apprenticeships data) back just to retrieve the AccountLegalEntityId. Second is that this endpoint could be used by the EI-40 story, to get the AccountLegalEntityID if it's not supplied in the URL

Pauljgraham: I can see you're logic here you only need one field. Should the method be called 'GetApplicationAccountLegalEntityId' as that is what it's doing. Or are we planning on adding the AccountLegalEntityId and the LegalEntityId to this call in the future

src/SFA.DAS.EmployerIncentives.Api/Models/SendBankDetailsEmailRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿
+namespace SFA.DAS.EmployerIncentives.Api.Models
+{
+    public class SendBankDetailsEmailRequest
+    {
+        public long AccountId { get; set; }
+        public long AccountLegalEntityId { get; set; }
```
Pauljgraham, line 7: I may well I have slightly misunderstood the purpose of this functionality, but don't we send the bank the LegalEntityId not the accountLegalEntityId. However I see it's for an email so it's probably correct

src/SFA.DAS.EmployerIncentives/Application/Services/EmployerIncentivesService.cs
David-B-Read: I've renamed in this branch to hopefully avoid confusion:
https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/85/commits/cccb522ece4707229fb5f0efe26b9937780e5906

src/SFA.DAS.EmployerIncentives.Api/Models/SendBankDetailsEmailRequest.cs
David-B-Read: It's for monitoring purposes, if we have a failed message then we know which account/account legal entity it's for

David-B-Read:
Merged as part of https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/85

PR #75, opened 2020-08-11
Build/Release YAML

adam230594:
- Deploys to AT, TEST, TEST2 and PP
- Deploys FAT, Reservations, Employer Incentives and Find Epao outer API's
- Uses tags of das-platform-automation and das-platform-building-blocks for backwards compatibility
- There is no triggers for CI.


azure-pipelines-release.yml
```diff
@@ -0,0 +1,730 @@
+parameters:
```
NickGraham101, line 1: if this file replaces azure-pipelines.yml then azure-pipelines.yml should be deleted as part of this PR

PR #86, opened 2020-08-17
EI-215/208 get agreement info

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives/InnerApi/Requests/Accounts/GetAccountLegalEntityRequest.cs
```diff
@@ -0,0 +1,21 @@
+﻿using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.EmployerIncentives.InnerApi.Requests.Accounts
+{
+    public class GetAccountLegalEntityRequest : IGetApiRequest
```
David-B-Read, line 5: Could this class be renamed? This request gets the "legal entity id", there's already a request defined in other branches (EI-40, EI-145, EI-210) to get the "account legal entity id", could get even more confusing than the fact that there are two ids with similar names :)

ruslanss (approved):

dashton82 (approved):

PR #88, opened 2020-08-18
Fat2 129 use lars data for duration and funding master target

PeterDavidWilliams:


src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Responses/GetStandardsListItem.cs
```diff
@@ -11,14 +11,14 @@ public class GetStandardsListItem
...
+        [JsonIgnore]
+        public long MaxFunding => GetFundingDetails("MaxFunding");
 
         public string OverviewOfRole { get; set; }
 
         public string Keywords { get; set; }
-
-        public int TypicalDuration { get; set; }
+        [JsonIgnore]
+        public int TypicalDuration => (int)GetFundingDetails("TypicalDuration");
```
dashton82, line 21:
```suggestion
        public int TypicalDuration => (int)GetFundingDetails(nameof(TypicalDuration));
```

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Responses/GetStandardsListItem.cs
```diff
@@ -11,14 +11,14 @@ public class GetStandardsListItem
         public string Title { get; set; }
         public int Level { get; set; }
         public decimal Version { get; set; }
-        [JsonIgnore] 
-        public long MaxFunding => GetMaxFunding();
+        [JsonIgnore]
+        public long MaxFunding => GetFundingDetails("MaxFunding");
```
dashton82, line 15:
```suggestion
        public long MaxFunding => GetFundingDetails(nameof(MaxFunding));
```

Not quite sure why this is a long? would make sense for it to be an int as thats whats coming from the inner API

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Responses/GetStandardsListItem.cs
```diff
@@ -31,39 +31,67 @@ public class GetStandardsListItem
...
+
+        private long GetFundingDetails(string prop)
         {
             var funding = ApprenticeshipFunding
-                .FirstOrDefault(c => 
-                    c.EffectiveFrom <= DateTime.UtcNow && (c.EffectiveTo == null 
+                .FirstOrDefault(c =>
+                    c.EffectiveFrom <= DateTime.UtcNow && (c.EffectiveTo == null
                                                            || c.EffectiveTo >= DateTime.UtcNow));
+            if (prop == "MaxFunding")
```
dashton82, line 41: i think you can probably keep all of this in one method, there is duplication in those two methods where all it is trying to do is to find a ApprenticeshipFunding item that meets the criteria

dashton82 (approved):

PR #101, opened 2020-08-24
FAT2-152 location search

dashton82:
Added location controller to get locations from locations api 🤔 ....location

### DIFF
--- a/src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetProvidersByCourseRequest.cs
+++ b/src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetProvidersByCourseRequest.cs
@@ -6,16 +6,18 @@ namespace SFA.DAS.FindApprenticeshipTraining.InnerApi.Requests
     {
         private readonly double? _latitude;
         private readonly double? _longitude;
+        private readonly int _sortOrder;
         private readonly int _courseId;
         
-        public GetProvidersByCourseRequest (int id, double? latitude = null, double? longitude = null)
+        public GetProvidersByCourseRequest (  int id, double? latitude = null, double? longitude = null, int sortOrder = 0)
         {
             _latitude = latitude;
             _longitude = longitude;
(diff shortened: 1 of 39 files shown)


src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetProvidersByCourseRequest.cs
```diff
@@ -4,14 +4,18 @@ namespace SFA.DAS.FindApprenticeshipTraining.InnerApi.Requests
 {
     public class GetProvidersByCourseRequest : IGetApiRequest
     {
-        private int _courseId { get ; set ; }
+        private readonly double? _latitude;
+        private readonly double? _longitude;
+        private int CourseId { get ;}
```
PeterDavidWilliams, line 9: Is there a reason this was changed from `_courseId`?

src/SFA.DAS.FindApprenticeshipTraining/Services/LocationApiClient.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Collections.Generic;
```
PeterDavidWilliams, line 1: I'm assuming that all the `NotImplemented` is because we don't need those calls right now?

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetProvidersByCourseRequest.cs
dashton82: Naming convention, if its readonly then it should be `_courseId`. However its a Get only property at the moment so should be named `CourseId`

Having said that, i'll push up a change to make it consistent with the other properties

PeterDavidWilliams (approved):

PeterDavidWilliams (approved):

PeterDavidWilliams (approved):

scottywakefield (approved):

PR #117, opened 2020-09-02
Add-named-client-to-throttle-calls

Stephek:


David-B-Read (approved):

David-B-Read (approved):

src/SFA.DAS.EmployerIncentives.Api/AppStart/AddServiceRegistrationExtension.cs
```diff
@@ -23,6 +26,12 @@ public static void AddServiceRegistration(this IServiceCollection services)
             services.AddTransient<IEmployerIncentivesService, EmployerIncentivesService>();
             services.AddTransient<ICommitmentsService, CommitmentsService>();
             services.AddTransient<IAccountsService, AccountsService>();
+
+            int maxParallelism = 100;
+            var throttler = Policy.BulkheadAsync<HttpResponseMessage>(maxParallelism, int.MaxValue);
+
+            services.AddHttpClient("Throttled")
+                .AddPolicyHandler(throttler);
```
dashton82, line 34: perhaps link this to the Client that you want this applying to:

```
int maxParallelism = 100;
            var throttler = Policy.BulkheadAsync<HttpResponseMessage>(maxParallelism, int.MaxValue);

            services.AddHttpClient(nameof(EmployerIncentivesConfiguration))
                .AddPolicyHandler(throttler);
```

And then change the APIClient to the below 👇🏻

src/SFA.DAS.SharedOuterApi/Infrastructure/ApiClient.cs
```diff
@@ -24,20 +24,18 @@ public class ApiClient<T> : IApiClient<T> where T : IInnerApiConfiguration
             IWebHostEnvironment hostingEnvironment,
             IAzureClientCredentialHelper azureClientCredentialHelper)
         {
-            _httpClient = httpClientFactory.CreateClient();
+            _httpClientFactory = httpClientFactory;
```
dashton82, line 27:
```
_httpClient = httpClientFactory.CreateClient(typeof(T).Name);
```
This links to the above, having them all named causes no issues and will also mean that all of the other changes aren't necessary. Its more likely that the throttling would need to be done across every client

dashton82 (changes requested):
I've remove the `Throttled` client name and instead have every client have the option to apply its own http configuration.

PR #119, opened 2020-09-02
Combine apprenticeship eligibility checks into a single inner API call

David-B-Read:


MichaelYoung1981:
A GET that does a POST 🤯

David-B-Read:
> A GET that does a POST 🤯

The Outer API contract should still be a GET, from the client's perspective. Posting JSON data in the request body of a GET request isn't officially supported, so that's why the inner API request type was changed to a POST, so that the data from commitments API can be passed over in one go

David-B-Read:
Closing as re-submitting under branch with ticket number

PR #136, opened 2020-09-14
FAT2-166 - Add location property to providers response

dashton82:
* Return default if not found from get request


dashton82:
This fix has gone into master but not into sprint_release somehow 🤔

scottywakefield (approved):

PR #153, opened 2020-09-28
FAT2-220 - Fix Location Search For Providers Bug

nouriach:
Create new request for when a user searches for just a location name


SijiOdun:
PR not required for fix

PR #157, opened 2020-10-01
FAT2-206 filter providers by rating

dashton82:


src/SFA.DAS.FindApprenticeshipTraining.Api.UnitTests/Models/WhenCastingToProviderBaseTypesFromMediatorType.cs
```diff
@@ -403,9 +403,56 @@ public void The_If_There_Are_Delivery_Modes_To_Filter_And_Returns_No_Delivery_Mo
...
+                new GetFeedbackRatingItem
+                {
+                    FeedbackName = "Excellent",
+                    FeedbackCount = 1,
+                }
+            };
+            
+            var response = new GetTrainingCourseProviderListItem().Map(source, sectorSubjectArea,1,new List<DeliveryModeType>(), new List<FeedbackRatingType>{FeedbackRatingType.Poor,FeedbackRatingType.Excellent});
+
+            response.Should().NotBeNull();
```
PeterDavidWilliams, line 454: Can you modify this so that instead of it just not checking for null, it checks for what it actually should be

PeterDavidWilliams (changes requested):

scottywakefield (approved):

PeterDavidWilliams (approved):

scottywakefield (approved):

PR #173, opened 2020-10-13
Build/Release Yaml

adam230594:
- Deploys to AT, TEST, TEST2 and PP
- Deploys FAT, Reservations, Employer Incentives and Find Epao outer API's
- Uses tags of das-platform-automation and das-platform-building-blocks for backwards compatibility
- There is no triggers for CI.


jack-education (approved):

azure-pipelines-release.yml
```diff
@@ -0,0 +1,730 @@
...
+  - repository: das-platform-automation
+    type: github
+    name: SkillsFundingAgency/das-platform-automation
+    ref: refs/tags/4.2.0
+    endpoint: GitHub (SFA)
+  pipelines:
+  - pipeline: das-employer-config
+    project: Digital Apprenticeship Service
+    source: das-employer-config
+    branch: master
```
NickGraham101, line 40: Should this take a parameter (that defaults to master) to allow a branch of config to be used (as we do with most classic releases)?

azure-pipelines-release.yml
```diff
@@ -0,0 +1,730 @@
...
+    environment: DEV
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+          - template: azure-pipelines-templates/deploy/step/app-deploy.yml@das-platform-building-blocks
+            parameters:
+              ServiceConnection: SFA-DAS-DevTest-ARM
+              AppServiceName: $(FindApprenticeshipOuterApiAppServiceName)
+              DeploymentPackagePath: $(Pipeline.Workspace)/ApimEndpointsArtifacts/SFA.DAS.FindApprenticeshipTraining.Api.Zip
```
NickGraham101, line 149: The filetype suffix .zip is normally lowercase.  If the deployment package is being published with a lowercase suffix this task will fail on a linux agent (this issue reoccurs another 15 times :) )

azure-pipelines-release.yml
```diff
@@ -0,0 +1,730 @@
...
+      TemplateSecrets:
+        LoggingRedisConnectionString: $(LoggingRedisConnectionString)
+        ConfigurationStorageConnectionString: $(ConfigurationStorageConnectionString)
+
+  - deployment: Deploy_FindApprenticeshipTraining
+    dependsOn: DeployInfrastructure
+    condition: and(succeeded(), eq(${{ parameters.FindApprenticeshipTraining }}, true))
+    variables:
+      ResourceGroupName: $[dependencies.DeployInfrastructure.outputs['DeployInfrastructure.ArmOutputs.ResourceGroupName']]
+    environment: DEV
```
NickGraham101, line 75: Should the environment be called DEVTEST (or maybe DevTest) to be consistent with the way these environments (AT, TEST, PP) are generally referred to?

adam230594: DevTest should refer to our subscription that deploys to AT TEST TEST2 and DEMO but yeah we do generally include PP when we refer to DevTest (to make matters more confusing!)

Going forward we could correct this and have a DEV environment to deploy to AT TEST TEST2 PP.
PROD -> Production stages
MO -> MO stages

NickGraham101: Sounds reasonable

azure-pipelines-release.yml
adam230594: Not sure if its possible to pass a pipeline branch input in as a parameter as I'm getting YAML failures.

Its one of the kinks of YAML release pipelines that hasn't been implemented yet https://developercommunity.visualstudio.com/content/problem/1064010/yaml-pipelines-variable-pipeline-resources.html

NickGraham101: I should've checked the docs, this is possible out of the box without any template changes - https://docs.microsoft.com/en-us/azure/devops/pipelines/process/resources?view=azure-devops&tabs=schema#manual-version-picker-for-resources-in-the-create-run-dialogue

azure-pipelines-release.yml
```diff
@@ -0,0 +1,409 @@
...
+  - template: azure-pipelines-templates/job/deploy-api.yml
+    parameters:
+      DeploymentName: Deploy_FindApprenticeshipTraining
+      DeploymentCondition: and(succeeded(), eq(${{ parameters.FindApprenticeshipTraining }}, true))
+      DeploymentApprovals: DEV
+      ServiceConnection: SFA-DAS-DevTest-ARM
+      AppServiceName: $(FindApprenticeshipOuterApiAppServiceName)
+      DeploymentPackagePath: $(Pipeline.Workspace)/ApimEndpointsArtifacts/SFA.DAS.FindApprenticeshipTraining.Api.zip
+      ApimResourceGroup: $(ApimResourceGroup)
+      InstanceName: $(InstanceName)
```
NickGraham101, line 79: ApimInstanceName would be clearer

azure-pipelines-templates/job/deploy-api.yml
```diff
@@ -0,0 +1,49 @@
+parameters:
+  DeploymentName: ''
```
NickGraham101, line 2: template parameters default values should be left blank unless the template is designed to succeed with an empty string as the value for a particular parameter

NickGraham101: So `DeploymentCondition: false` is good but the other parameters appear to be required params rather than optional so should be `DeploymentName: ` rather than `DeploymentName: ''`

adam230594: oooo, I'll get them changed in the platform-building blocks as well then as they also have an empty string as the default when it shouldnt

NickGraham101 (approved):

PR #176, opened 2020-10-16
EI-279/ EI-501 Limit query to daily results

ruslanss:


MichaelYoung1981 (approved):

src/SFA.DAS.EmployerIncentives/Application/Commands/UpdateVendorRegistrationFormCaseStatus/RefreshVendorRegistrationFormCaseStatusCommand.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace SFA.DAS.EmployerIncentives.Application.Commands.UpdateVendorRegistrationFormCaseStatus
 {
-    public class RefreshVendorRegistrationFormCaseStatusCommand : IRequest
+    public class RefreshVendorRegistrationFormCaseStatusCommand : IRequest<DateTime>
```
David-B-Read, line 6: Possibly over-engineering, but would it be better to return a Response type rather than a primitive type? I've seen that pattern elsewhere...

David-B-Read (approved):

MichaelYoung1981 (dismissed):

src/SFA.DAS.EmployerIncentives/Application/Commands/UpdateVendorRegistrationFormCaseStatus/RefreshVendorRegistrationFormCaseStatusCommand.cs
ruslanss: Hi David, there is a tech debt Jira ticket to rework this whole thing :)

MichaelYoung1981 (approved):

PR #178, opened 2020-10-20
FAT2-207 get course epaos

scottywakefield:


src/SFA.DAS.SharedOuterApi/Interfaces/IAssessorsApiClient.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace SFA.DAS.SharedOuterApi.Interfaces
+{
+    public interface IAssessorsApiClient<T> : IInternalApiClient<T>
```
dashton82, line 3: There is an `IGetApiClient<T>` been added which might be better for this?

scottywakefield: yeah will do.

dashton82 (approved):

dashton82 (dismissed):

dashton82 (approved):

PR #199, opened 2020-11-11
Fat2 243 order job titles

PeterDavidWilliams:


src/SFA.DAS.FindApprenticeshipTraining.Api/Models/GetTrainingCourseListItem.cs
```diff
@@ -49,6 +50,16 @@ public class GetTrainingCourseListItem
...
         }
+        private static string OrderJobTitles(string jobTitles)
+        {
+            var titles = jobTitles.Split(',').OrderBy(x => x);
+            var titleString = "";
+            foreach (var title in titles)
+            {
+                titleString += title + ",";
+            }
+            return titleString.TrimEnd(',');
```
scottywakefield, line 61: this can be done with string.Join() and will be more efficient than lots of string concat operations.

scottywakefield (changes requested):

scottywakefield (approved):

src/SFA.DAS.FindApprenticeshipTraining.Api/Models/GetTrainingCourseListItem.cs
PeterDavidWilliams: Going back in to make this change

scottywakefield (approved):

PR #213, opened 2020-11-30
EI-610 set active collection calendar period

David-B-Read:


ruslanss:
@Stephek had a good idea for a database constraint to only allow 1 active calendar period record. do you think we can add it as part of this story?

David-B-Read:
> @Stephek had a good idea for a database constraint to only allow 1 active calendar period record. do you think we can add it as part of this story?

The inner API enforces only one active collection calendar period

src/SFA.DAS.EmployerIncentives.Api/Controllers/CollectionCalendarController.cs
```diff
@@ -0,0 +1,32 @@
...
+        private readonly ILogger<CollectionCalendarController> _logger;
+
+        public CollectionCalendarController(IMediator mediator, ILogger<CollectionCalendarController> logger)
+        {
+            _mediator = mediator;
+            _logger = logger;
+        }
+
+        [HttpPatch]
+        [Route("/collectionCalendar/period/active")]
```
Stephek, line 22: Shouldn't this be /collectionPeriods and the patch is just the CollectionPeriod data to be updated

src/SFA.DAS.EmployerIncentives.Api/Models/ActivateCollectionCalendarPeriodRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace SFA.DAS.EmployerIncentives.Api.Models
+{
+    public class ActivateCollectionCalendarPeriodRequest
```
Stephek, line 4: I think this is just a CollectionPeriodRequest with Number, Year and IsActive.  It doesn't make sense to call it Activate... when the Active field could be false.

Stephek (changes requested):

Stephek (dismissed):

ruslanss (approved):

PR #214, opened 2020-12-01
FAT2-287 epao register api (decommish fat v1 in DC)

scottywakefield:


src/SFA.DAS.EpaoRegister/Application/Epaos/Queries/GetEpaoCourses/GetEpaoCoursesQueryHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+            
+            var courseTasks = new List<Task<GetStandardResponse>>();
+            foreach (var epaoCoursesListItem in epaoCoursesListItems)
+            {
+                var getCourseTask = _coursesApiClient.Get<GetStandardResponse>(
+                    new GetStandardRequest(epaoCoursesListItem.StandardCode));
+                courseTasks.Add(getCourseTask);
+            }
+            await Task.WhenAll(courseTasks);
```
dashton82, line 56: not sure if its worth getting all the courses
``` 
var coursesTask = _cacheHelper.GetRequest<GetStandardsListResponse>(_coursesApiClient,
                new GetStandardsListRequest(), nameof(GetStandardsListResponse), out var saveToCache);
```
and saving to cache - then iterating over the cached list for the courses that you want

src/SFA.DAS.EpaoRegister/InnerApi/Requests/GetEpaoRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.EpaoRegister.InnerApi.Requests
+{
+    public class GetEpaoRequest : IGetApiRequest
+    {
+        public string EpaoId { get; set; }
```
dashton82, line 7: not sure we do everywhere but it would probably be better to have this set in the constructor

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetStandardRequest.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace SFA.DAS.FindApprenticeshipTraining.InnerApi.Requests
 {
-    public class GetStandardRequest : IGetApiRequest
+    public class GetStandardRequest : IGetApiRequest // todo: remove once merged up to date
```
dashton82, line 5: Can that be removed now?

src/SFA.DAS.EpaoRegister/InnerApi/Requests/GetEpaoCoursesRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.EpaoRegister.InnerApi.Requests
+{
+    public class GetEpaoCoursesRequest : IGetAllApiRequest
+    {
+        public string EpaoId { get; set; }
+        public string GetAllUrl => $"api/ao/assessment-organisations/{EpaoId}/standards";
```
dashton82, line 8: not sure we do everywhere but it would probably be better to have this set in the constructor

dashton82 (changes requested):

src/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetStandardRequest.cs
scottywakefield: lol yes

dashton82 (approved):

PeterDavidWilliams (approved):

PR #224, opened 2020-12-21
Ei 733 get vendor id on vrf status update

David-B-Read:


src/SFA.DAS.EmployerIncentives/Application/Commands/UpdateVendorRegistrationFormCaseStatus/RefreshVendorRegistrationFormCaseStatusCommandHandler.cs
```diff
@@ -48,25 +52,31 @@ public async Task<DateTime> Handle(RefreshVendorRegistrationFormCaseStatusComman
...
 
             await Task.WhenAll(response.RegistrationCases.Select(UpdateVendorRegistrationCaseStatus));
 
             return await Task.FromResult(nextRunDateTime);
         }
 
+        private async Task UpdateVendorRegistrationCaseStatus(VendorRegistrationCase @case)
+        {
+            var accountLegalEntity = await _incentivesService.GetLegalEntityByHashedId(@case.ApprenticeshipLegalEntityId);
+            if (String.IsNullOrWhiteSpace(accountLegalEntity?.VrfVendorId))
```
ruslanss, line 65: they send "000000" i believe

src/SFA.DAS.EmployerIncentives/Application/Commands/UpdateVendorRegistrationFormCaseStatus/RefreshVendorRegistrationFormCaseStatusCommandHandler.cs
```diff
@@ -48,25 +52,31 @@ public async Task<DateTime> Handle(RefreshVendorRegistrationFormCaseStatusComman
...
+            await _incentivesService.UpdateVendorRegistrationCaseStatus(
+                new UpdateVendorRegistrationCaseStatusRequest
+                {
+                    CaseId = @case.CaseId,
+                    HashedLegalEntityId = @case.ApprenticeshipLegalEntityId,
+                    Status = @case.CaseStatus,
+                    CaseStatusLastUpdatedDate = @case.CaseStatusLastUpdatedDate
+                });
+        }
+
```
MichaelYoung1981, line 79: Wouldn't it be better to use the domain event approach we use when the case is Completed?  That way there would be a single way of doing it.

David-B-Read: Is that something that could be picked up as part of the planned refactoring?

PR #235, opened 2021-01-14
Fat2 363 epao availiability

PeterDavidWilliams:


src/SFA.DAS.FindEpao.Api/Models/GetCourseEpaoListItem.cs
```diff
@@ -9,6 +10,7 @@ public class GetCourseEpaoListItem
         public string Name { get; set; }
         public string City { get; set; }
         public string Postcode { get; set; }
+        public DateTime? EffectiveFrom { get; set; }
```
dashton82, line 13: Not convinced this should be nullable

dashton82 (changes requested):

src/SFA.DAS.FindEpao.Api/Models/GetCourseEpaoListItem.cs
PeterDavidWilliams: change made

PeterDavidWilliams: Change made

dashton82 (approved):

PR #245, opened 2021-01-21
Cs 130 added acceptance tests 

Pauljgraham:


src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Bindings/InnerApi.cs
```diff
@@ -0,0 +1,23 @@
...
+        public InnerApi(TestContext context)
+        {
+            _context = context;
+        }
+
+        [BeforeScenario(Order = 1)]
+        public void Initialise()
+        {
+            if(_context.InnerApi == null)
+                _context.InnerApi = new MockApi();
```
benagain, line 20: It does't matter one bit, but there's funky new(ish) syntax:

    public void Initialise()
    {
        _context.InnterApi ??= new MockApi();
    }
`

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Features/HealthChecks.feature
```diff
@@ -0,0 +1,20 @@
...
+
+Scenario Outline: HealthCheck is called
+	Given the Apprentice Commitments Inner Api is ready and <InnerStatus>
+	When I call the health endpoint of the Outer Api
+	Then the result should show <OuterStatus>
+
+Examples:
+	| InnerStatus | OuterStatus |
+	| Healthy     | Healthy     |
+	| Unhealthy   | Unhealthy   |
```
benagain, line 20: `InnerStatus` will *always* be the same as `OuterStatus` - I wonder if it is really worth having separate columns.  Maybe a single `Status` column will help express that we are expecting them to be identical?

src/SFA.DAS.ApprenticeCommitments.Api/Controllers/ApprenticeshipController.cs
```diff
@@ -17,7 +17,7 @@ public class ApprenticeshipController : ControllerBase
         public async Task<IActionResult> AddApprenticeship(CreateApprenticeshipCommand request)
         {
             await _mediator.Send(request);
-            return Ok();
+            return Accepted();
```
benagain, line 20: I agree we should change this.  I'm not sure it belongs in a PR about adding acceptance tests?  That makes it hard if someone comes along later to discover why this change was made.  Maybe it would be better to undo this change for this particular PR and create a separate, very simple PR to change the behaviour.

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Bindings/InnerApi.cs
Pauljgraham: I like it

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Features/HealthChecks.feature
Pauljgraham: It will be different when we have multiple APIs (Notifications/Commitments etc)

benagain (approved):

PR #266, opened 2021-01-28
Add email template fields for new apprenticeship invitation

benagain:


src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
```diff
@@ -1,27 +1,51 @@
...
+            CancellationToken cancellationToken)
         {
             var id = Guid.NewGuid();
-            await _apprenticeCommitmentsService.CreateApprenticeship(id, command.ApprenticeshipId, command.Email);
-            await _apprenticeLoginService.SendInvitation(id, command.Email);
+
+            var apprentice = await _commitmentsService.GetApprenticeshipDetails(
+                command.EmployerAccountId,
+                command.ApprenticeshipId);
+
```
Pauljgraham, line 34: Can we verify that we should get the apprenticeship name from comitments/approvals. I don't mind doing it (in fact I can see some good reasons to do it), but it's not the source of truth for that data element.  And from memory commitments does call FAT to get it (but I can't remember if this is then stored)

src/SFA.DAS.ApprenticeCommitments/Application/Services/CommitmentsV2Service.cs
```diff
@@ -0,0 +1,32 @@
...
+                new GetApprenticeshipDetailsRequest(apprenticeshipId));
+
+            if (apprenticeship?.EmployerAccountId != accountId)
+            {
+                throw new HttpRequestContentException(
+                    $"Employer Account {accountId} does not have access to apprenticeship Id {apprenticeshipId}",
+                    System.Net.HttpStatusCode.BadRequest,
+                    "");
+            }
+
```
Pauljgraham, line 28: Good check

Pauljgraham (commented):
Just a small comment on where we get the course details from

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Steps/AddApprenticeshipSteps.cs
```diff
@@ -76,6 +79,27 @@ public void GivenTheApprenticeLoginApiIsReady()
                 );
         }
 
+        [Given(@"the Approvals API is ready")]
+        public void GivenTheApprovalsAPIIsReady()
+        {
+            _context.CommitmentsV2InnerApi.MockServer
+                .Given(
+                    Request.Create().WithPath($"/api/apprenticeships/1020")
```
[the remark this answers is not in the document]
benagain: Good spot.

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
benagain: @roysbailey - What do you think?  On the one hand, we are already calling the Commitments API and it returns the course name.  On the other hand it is not the source of truth.  This job will be running soon after the approval event, so one would imagine the Commitments API is up-to-date at that point, maybe?

PR #278, opened 2021-02-08
Uncomment for deployment to prod for Epao Register

dashton82:


cpwevans (commented):
Is there a reason why you have uncommented the PRD deployment of Reservations and ManageApprenticeships when  they haven't been deployed to PRD before and you are only wanting to deploy EpaoReg to PRD?

dashton82:
We had an issue earlier in the week where changes had gone in to master after the point that we had tested up-to and wanted to release. Subsequently if we had pushed the yml changes into master we would have deployed un-tested code for that outer api. To mitigate this issue it makes sense to have all of the environments in the yml so that master is always deployable to each env at that point in time

cpwevans (approved):

PR #293, opened 2021-02-22
SV-444 postman tests for id changes

DevenKShah:


CodeWolf86:
This looks good Deven, I can see that we use the get on Courses Api with the two seperate filters to confirm the counts coming back for each of the Outer Apis. Can also see we confirm that each of them contain the respective first standard that's returned from the Active Available filter.

One thing to consider though it doesn't necessarily stop this merge going in:

The Ifate Api is open: https://www.instituteforapprenticeships.org/api/apprenticeshipstandards
If we wanted to be more concrete, we could have a 2 stage test. One for IFate to Courses Api, One for Courses Api to Outer Apis

If we added a Get Request to the above, we could filter the response and re-enact some of the filters put in the Course Api to confirm we get the right values out e.g.:

`var jsonData = pm.response.json();
var approvedStandards = jsonData.filter(obj => obj.status == 'Approved for delivery');
console.log(approvedStandards.length);`

As you say the data does change over time, but by getting some of the data to assert from the source it could make the tests really tight. What do you think?

CodeWolf86 (approved):

DevenKShah:
> This looks good Deven, I can see that we use the get on Courses Api with the two seperate filters to confirm the counts coming back for each of the Outer Apis. Can also see we confirm that each of them contain the respective first standard that's returned from the Active Available filter.
> 
> One thing to consider though it doesn't necessarily stop this merge going in:
> 
> The Ifate Api is open: https://www.instituteforapprenticeships.org/api/apprenticeshipstandards
> If we wanted to be more concrete, we could have a 2 stage test. One for IFate to Courses Api, One for Courses Api to Outer Apis
> 
> If we added a Get Request to the above, we could filter the response and re-enact some of the filters put in the Course Api to confirm we get the right values out e.g.:
> 
> `var jsonData = pm.response.json(); var approvedStandards = jsonData.filter(obj => obj.status == 'Approved for delivery'); console.log(approvedStandards.length);`
> 
> As you say the data does change over time, but by getting some of the data to assert from the source it could make the tests really tight. What do you think?

Good shout, shouldn't Outer api be agnostic to where data is coming from in courses api? However, I think this strategy is good for writing tests on Courses API.

PR #299, opened 2021-02-25
SV-467 assessor service to return all standards

DevenKShah:


CodeWolf86 (approved):
All good, Just had to amend one of the Fat UI Requests to fit in so might be worth a check on my commit 👍

PR #300, opened 2021-02-25
SV-170 Changes for standards versioning

DevenKShah:
We have renamed Id  to LarsCode in Courses API response, this is mainly to map all outer response's Id to LarsCode.


src/SFA.DAS.FindEpao.Api.UnitTests/Controllers/Courses/WhenGettingCourseList.cs
```diff
@@ -35,7 +36,10 @@ public class WhenGettingCourseList
             controllerResult.StatusCode.Should().Be((int)HttpStatusCode.OK);
             var model = controllerResult.Value as GetCourseListResponse;
             Assert.IsNotNull(model);
-            model.Courses.Should().BeEquivalentTo(mediatorResult.Courses);
+            model.Courses.Should().BeEquivalentTo(mediatorResult.Courses, options => options
+            .Excluding(c => c.StandardUId)
+            .Excluding(c => c.LarsCode));
+            model.Courses.All(c => c.Id > 0).Should().BeTrue();
```
scottywakefield, line 42: if there's already a test covering the casting of types, just assert that the casts are equivalent, then you don't have to worry about all Id > 0 (they could all be 1 and pass this). e.g. 
`model.Courses.Should().BeEquivalentTo(mediatorResult.Courses.Select(course => (TypeCastTo)course)`

src/SFA.DAS.Reservations.Api.UnitTests/Controllers/WhenCallingGetTrainingCourseList.cs
```diff
@@ -36,7 +37,10 @@ public class WhenCallingGetTrainingCourseList
             controllerResult.StatusCode.Should().Be((int)HttpStatusCode.OK);
             var model = controllerResult.Value as GetTrainingCoursesListResponse;
             Assert.IsNotNull(model);
-            model.Standards.Should().BeEquivalentTo(mediatorResult.Courses);
+            model.Standards.Should().BeEquivalentTo(mediatorResult.Courses, options => options
+                .Excluding(s => s.StandardUId)
+                .Excluding(s => s.LarsCode));
+            model.Standards.All(m => m.Id > 0).Should().BeTrue();
```
scottywakefield, line 43: if there's already a test covering the casting of types, just assert that the casts are equivalent, then you don't have to worry about all Id > 0 (they could all be 1 and pass this). e.g. 
`model.Courses.Should().BeEquivalentTo(mediatorResult.Courses.Select(course => (TypeCastTo)course)`

scottywakefield (changes requested):

src/SFA.DAS.FindEpao.Api.UnitTests/Controllers/Courses/WhenGettingCourseList.cs
DevenKShah: There is no equivalent property `Id` in the other object. It maps to `LarsCode` instead of `Id`. Hence we are just making sure here that  the `Id` in outer object is populated.

DevenKShah: Fixed

src/SFA.DAS.Reservations.Api.UnitTests/Controllers/WhenCallingGetTrainingCourseList.cs
DevenKShah: Fixed

scottywakefield (approved):

PR #305, opened 2021-03-05
Apprentice Commitments provides get current apprenticeship

benagain:


src/SFA.DAS.ApprenticeCommitments.Api/Controllers/ListApprenticeshipsController.cs
```diff
@@ -0,0 +1,31 @@
...
+        [HttpGet("/apprentices/{apprenticeId}/apprenticeships")]
+        public async Task<IActionResult> AddApprenticeship(Guid apprenticeId)
+        {
+            var response = await _client.Get<GetApprenticeshipsRepsonse>(
+                            new GetApprenticeshipsRequest(apprenticeId));
+            if (response == null)
+                return NotFound();
+            else
+                return Ok(response);
+        }
```
Pauljgraham, line 29: One of the patterns people complain to me about is 'if/else' statements should be bracketed '{}' but I don't mind wither way. Not sure if it's really a ESFA standard or just a personal preference for those

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Features/GetApprenticeship.feature
```diff
@@ -0,0 +1,25 @@
...
+Scenario: Apprenticeship overview
+	Given there is an apprenticeship
+	When the apprenticeship overview is requested
+	Then the result should be OK
+	And the apprenticeship is returned
+
+Scenario: No current apprenticeship
+	Given there is no apprenticeship
+	When the apprenticeship overview is requested
+	Then the result should be NotFound
```
Pauljgraham, line 24: This is not a concern for this situation, but when we ask for a list of things and it's empty. Is it not more common to return a response as an empty list? But in our case there should always be one apprenticeship.

Pauljgraham (approved):

Pauljgraham (approved):

PR #306, opened 2021-03-08
Change apprentice identifier to be a GUID

benagain:


src/SFA.DAS.ApprenticeCommitments.Api/Controllers/QueryForwardingController.cs
```diff
@@ -0,0 +1,35 @@
...
+        public QueryForwardingController(GenericInnerApiQueryClient client)
+        {
+            this.client = client;
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> ForwardToInnerApi()
+        {
+            var innerResponse = await client.GetAsync(HttpContext.Request.Path);
+
```
Pauljgraham, line 21: My first concern with this pattern is all the GETS to outer API will be passed to the inner API (this might be limiting).

src/SFA.DAS.ApprenticeCommitments.Api/Controllers/QueryForwardingController.cs
```diff
@@ -0,0 +1,35 @@
...
+                return NotFound();
+
+            var content = await innerResponse.Content.ReadAsStringAsync();
+
+            return new ContentResult
+            {
+                Content = content,
+                ContentType = "application/json",
+                StatusCode = 200
+            };
```
Pauljgraham, line 32: if acting as a passthrough client, then you would need to return the HttpCodeStatus from the inner api

src/SFA.DAS.ApprenticeCommitments.Api/Controllers/QueryForwardingController.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> ForwardToInnerApi()
+        {
+            var innerResponse = await client.GetAsync(HttpContext.Request.Path);
+
+            if (!innerResponse.IsSuccessStatusCode)
+                return NotFound();
+
```
Pauljgraham, line 24: Why not return the status code

src/SFA.DAS.ApprenticeCommitments.Api/Startup.cs
```diff
@@ -110,6 +111,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerF
             app.UseRouting();
             app.UseEndpoints(endpoints =>
             {
+                endpoints.MapFallbackToController(
+                    nameof(QueryForwardingController.ForwardToInnerApi),
+                    "QueryForwarding");
+
```
Pauljgraham, line 117: I'm intrigued how this works!

Pauljgraham (commented):
Sorry we can't do this. This is a cut down version of a Passthrough API client we wrote for the EI project. Which did the same job. The Passthrough API was reviewed by the APIM architecture group and rejected the idea (I really don't agree with this, but I don't think we can ignore this decision). 

The reason might be, we are directly exposing the inner API to the outside applications. I assume this pattern will allow any inner API endpoint to be called and this might be perceived as a security risk. 

This is a discussion we need to have with Roy to see the objections.

benagain:
Closing due to concern over the nature of pass-through.  Will discuss with architect community to find an acceptable way to minimise the number of wrapper classes that are required for simple API calls.

benagain:
Closing due to concern over the nature of pass-through. Will discuss with architect community to find an acceptable way to minimise the number of wrapper classes that are required for simple API calls.

PR #315, opened 2021-03-10
SV-554 Added get active standards endpoint in assessors

DevenKShah:


src/SFA.DAS.Assessors.Api/Controllers/TrainingCoursesController.cs
```diff
@@ -62,5 +62,27 @@ public async Task<IActionResult> GetByStandardUId(string standardUId)
                 return BadRequest();
             }
         }
+
+        [HttpGet]
+        [Route("active")]
```
CodeWolf86, line 67: So we have a get list to get everything, and now a /active to get just the active standards. Know that EPAO will want other ones like In Development, 

Are we happy that EPAO will essentially "subtract" between the two responses? And there isn't a future scenario where we want an additional filter like "in development"?

Essentially i'm debating whether this endpoint should be combined with the Get All with a filter, or left separate...

CodeWolf86 (changes requested):

src/SFA.DAS.Assessors.Api/Controllers/TrainingCoursesController.cs
```diff
@@ -62,5 +62,27 @@ public async Task<IActionResult> GetByStandardUId(string standardUId)
...
+                {
+                    Courses = queryResult.TrainingCourses.Select(c => (GetCourseListItem)c).ToList()
+                };
+
+                return Ok(model);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error attempting to get list of training courses");
+                return BadRequest();
```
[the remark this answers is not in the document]
DevenKShah: I didn't wanted to change this hear, as we are all changing the same file. I will (might already have) open a tech debt for later

src/SFA.DAS.Assessors.Api/Controllers/TrainingCoursesController.cs
DevenKShah: Valid point, I debated that myself. But because I don't know what the `in-development` requirement is, I would just go with this for now and when we have good idea about in-development, we can always refactor. What I didn't wanted to do is make assessor service aware of the filters, wanted to keep that with outer api.

CodeWolf86 (approved):

PR #320, opened 2021-03-12
Apprentice Commitments Confirm Your Training Provider

benagain:


src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Steps/AddApprenticeshipSteps.cs
```diff
@@ -113,6 +113,7 @@ public void ThenTheRequestToTheInnerApiWasMappedCorrectly()
             innerApiRequest.ApprenticeshipId.Should().Be(_request.ApprenticeshipId);
             innerApiRequest.EmployerName.Should().Be(_request.EmployerName);
             innerApiRequest.EmployerAccountLegalEntityId.Should().Be(_request.EmployerAccountLegalEntityId);
+            innerApiRequest.TrainingProviderName.Should().Be("Provisional Training Provider Name");
```
Pauljgraham, line 116: Just wondering if TrainingProviderId is supposed to be here?

src/SFA.DAS.ApprenticeCommitments/Apis/InnerApi/CreateApprenticeshipRequest.cs
```diff
@@ -17,6 +17,7 @@ public class CreateApprenticeshipRequestData
         public string Email { get; set; }
         public string EmployerName { get; set; }
         public long EmployerAccountLegalEntityId { get; set; }
+        public string TrainingProviderName { get; set; }
```
Pauljgraham, line 20: Definitely TrainingProviderId  should be here

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
```diff
@@ -40,7 +40,8 @@ public class CreateApprenticeshipCommandHandler : IRequestHandler<CreateApprenti
                 ApprenticeshipId = command.ApprenticeshipId,
                 Email = command.Email,
                 EmployerName = command.EmployerName,
-                EmployerAccountLegalEntityId = command.EmployerAccountLegalEntityId
+                EmployerAccountLegalEntityId = command.EmployerAccountLegalEntityId,
+                TrainingProviderName = "Provisional Training Provider Name",
             });
```
Pauljgraham, line 45: I'm wondering where are we getting the TrainingProviderId from. We actually have a choice of getting it from the event or from the call to commitments.

Pauljgraham (changes requested):
Think we are missing the TrainingProviderId

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
benagain: That will be part of CS-286

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Steps/AddApprenticeshipSteps.cs
benagain: Yep, I've added it in with a dummy value like the name.

src/SFA.DAS.ApprenticeCommitments/Apis/InnerApi/CreateApprenticeshipRequest.cs
benagain: Yep, I've added it in with a dummy value just like the name.

Pauljgraham (approved):

PR #324, opened 2021-03-16
CS-276 - Confirm your training provider - already confirmed

benagain:


MarieDale3 (dismissed):

src/SFA.DAS.ApprenticeCommitments/Apis/InnerApi/TrainingProviderConfirmationRequest.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public class TrainingProviderConfirmationRequestData
+    {
+        public bool TrainingProviderCorrect { get; set; }
+    }
+
+    public class TrainingProviderConfirmationResponse
+    {
+    }
```
Pauljgraham, line 34: There has to be a way to drop these awful empty classes

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Steps/ConfirmApprenticeshipSteps.cs
```diff
@@ -0,0 +1,62 @@
...
+            };
+        }
+
+        [Given("the inner API will accept the confirmation")]
+        public void GivenTheInnerAPIWillAcceptTheConfirmation()
+        {
+            _context.InnerApi.MockServer
+                .Given(
+                    Request.Create()
+                        .WithPath("/apprentices/*/apprenticeships/*/trainingproviderconfirmation")
```
Pauljgraham, line 41: My  concerns about the use of these wildcard matches are

1. Does it slow down the response from WireMock (I seem to see this but that anecdotal)
2. If we happen to pass in the wrong id in (or in the wrong position), our test will pass, but the code will fail

   Benefits are it's nicer than having to write explicit path for each case!

Pauljgraham (approved):
One tiny comment, feel free to ignore

Pauljgraham (approved):

PR #327, opened 2021-03-17
Cs 286 Added mock apis 

Pauljgraham:


benagain (approved):
There are more mock responses we can add for other scenarios but this is a great start 👍

PR #328, opened 2021-03-17
EI-1010-update-company-name-and-api-version

Stephek:


David-B-Read (changes requested):
We need to end to end test against the new API endpoint in OAT/UAT before the 6th April switchover. 
In 3 amigos/refinement we proposed putting the company name and API version into Azure config/pipeline deployment variables to enable this.

David-B-Read (approved):

PR #338, opened 2021-03-23
CS-221 Renamed RegistrationId to ApprenticeId, Implemented SendInvitationReminders with …

Pauljgraham:
…accept tests


src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommand.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using MediatR;
+
+namespace SFA.DAS.ApprenticeCommitments.Application.Commands.SendInvitationReminders
+{
+    public class SendInvitationRemindersCommand : IRequest
+    {
+        public DateTime SendNow { set; get; }
+        public int RemindAfterDays { set; get; }
```
benagain, line 9: I can't find where `RemindAfterDays` is used.  Should it be passed through to the inner API?

src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommandHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                _logger.LogInformation($"Sending Invitation for Apprentice {registration.ApprenticeId}");
+                await _apprenticeLoginService.SendInvitation(new SendInvitationModel
+                {
+                    SourceId = registration.ApprenticeId,
+                    Email = registration.Email,
+                    GivenName = registration.FirstName,
+                    FamilyName = registration.LastName,
+                    OrganisationName = registration.EmployerName,
+                    ApprenticeshipName = apprenticeship.CourseName
```
benagain, line 64: At some point I suspect we will store this in the Apprentice Commitments domain.  If we do we must remember to come back here and change this.

src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommandHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+            _logger.LogInformation("Getting Registrations which are due a reminder");
+            var list = (await _apprenticeCommitmentsService.GetReminderRegistrations()).Registrations;
+
+            // This just wouldn't work 
+            //Parallel.ForEach(list, new ParallelOptions {MaxDegreeOfParallelism = 4}, async registration =>
+            //{
+            //    await SendReminder(registration);
+            //});
+
+            await Task.WhenAll(list.Select(x=>SendReminder(x, command.SendNow)));
```
benagain, line 43: Is there a danger of bombarding the inner APIs and login service with a spike of HTTP requests if there are many apprenticeships that need reminding in a given day?  

Given a large enough number of reminders it sounds like the kind of thing where the first so many could succeed but then the latter part of the list fail.

src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommand.cs
Pauljgraham: oh dear, did I miss it from the query

src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommandHandler.cs
Pauljgraham: Agreed, but it depends on the Confirm your Apprenticeship story

src/SFA.DAS.ApprenticeCommitments/Application/Commands/SendInvitationReminders/SendInvitationRemindersCommandHandler.cs
Pauljgraham: This is also my concern. I worked a long time trying to user Parallel.ForEach() and throttling it but the call completed before the async tasks finished. It's worthwhile throttling this

benagain (dismissed):

benagain (approved):

benagain (dismissed):

benagain (dismissed):

benagain (approved):

PR #370, opened 2021-04-30
Clean up YAML deployments

adam230594:
Added arm-deploy.yml job template
Remove app-role-assignments.yml job template and add the step into arm-deploy.yml
Slimmed down azure-pipelines.yml by removing repeated parameters and referencing directly in job templates


pipeline-templates/job/arm-deploy.yml
```diff
@@ -0,0 +1,42 @@
+parameters:
+  ServiceConnection: ''
+  AppRoleAssignmentsServiceConnection: ''
+  Environment: ''
+
+jobs:
+- deployment: DeployInfrastructure
+  pool:
+    # name: DAS - Continuous Deployment Agents
+    name: DAS - Continuous Deployment
```
jack-cloud-platform, line 10: Does this need changing to the new agents?

pipeline-templates/job/deploy-api.yml
```diff
@@ -5,21 +5,19 @@ parameters:
...
   AppRoleAssignmentsServiceConnection:
-  Tenant:
 
 jobs:
 - deployment: ${{ parameters.DeploymentName }}
-  dependsOn: 
-  - ApplyAppRoleAssignments
+  pool:
+    # name: DAS - Continuous Deployment Agents
+    name: DAS - Continuous Deployment
```
jack-cloud-platform, line 19: Does this need changing to the new agents?

pipeline-templates/job/arm-deploy.yml
adam230594: Since the AKS agents have task cache issues, to not block the build/release split any longer we're going to run on the deployment vm

pipeline-templates/job/deploy-api.yml
adam230594: Since the AKS agents have task cache issues, to not block the build/release split any longer we're going to run on the deployment vm

jack-cloud-platform (approved):

PR #372, opened 2021-05-04
EI-973 add employment start date to apprenticeships

David-B-Read:
Add endpoint to update employment start dates


src/SFA.DAS.EmployerIncentives/Application/Commands/ConfirmEmploymentDetails/ConfirmEmploymentDetailsCommandHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+            var apprenticeshipDetails = new UpdateIncentiveApplicationRequestData
+            {
+                IncentiveApplicationId = command.ConfirmEmploymentDetailsRequest.ApplicationId,
+                AccountId = command.ConfirmEmploymentDetailsRequest.AccountId,
+                Apprenticeships = apprenticeships.Select(x => (IncentiveClaimApprenticeshipDto)x).ToArray()
+            };
+            foreach (var apprenticeship in apprenticeshipDetails.Apprenticeships)
+            {
+                var employmentDetails = command.ConfirmEmploymentDetailsRequest.EmploymentDetails.FirstOrDefault(x => x.ApprenticeId == apprenticeship.ApprenticeshipId);
+                if (employmentDetails != null)
```
ruslanss, line 36: don't think this null-check is required as the Apprenticeships list is driven by ID's in the EmploymentDetails ?

PR #376, opened 2021-05-10
Handle ApprenticeshipUpdatedApproved event

benagain:


Chazzysb (dismissed):
Approved

Pauljgraham (dismissed):

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/UpdateApprenticeshipCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+using SFA.DAS.ApprenticeCommitments.Application.Services;
+using SFA.DAS.ApprenticeCommitments.Application.Services.ApprenticeLogin;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using CommitmentsApprenticeshipResponse = SFA.DAS.ApprenticeCommitments.Apis.CommitmentsV2InnerApi.ApprenticeshipResponse;
+
+#nullable enable
+
+namespace SFA.DAS.ApprenticeCommitments.Application.Commands.CreateApprenticeship
```
Pauljgraham, line 13: sorry missed this earlier, the file is in the wrong folder. It should be in UpdateApprenticeship

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/UpdateApprenticeshipCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    public class UpdateApprenticeshipCommand : IRequest
+    {
+        public long ApprenticeshipId { get; set; }
+        public string Email { get; set; } = "";
+        public DateTime ApprovedOn { get; set; }
```
Pauljgraham, line 19: Email is not passed in from Jobs layer (and I don't think it's actually needed)

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/UpdateApprenticeshipCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var course = _coursesService.GetCourse(apprenticeship.CourseCode);
+            var provider = _trainingProviderService.GetTrainingProviderDetails(apprenticeship.ProviderId);
+
+            await Task.WhenAll(course, provider);
+
+            return (provider.Result, apprenticeship, course.Result);
+        }
+    }
+}
```
Pauljgraham, line 83: I still believe it's worthwhile putting these particular classes into individual files within the usual command structure. What happened here is, is we ended up putting this code in a file in the wrong folder and none of us spotted it first code review. 

Consider what happens when CI take over the project, they will quickly scan down the Command sections and make an overall assumption, that there is no UpdateApprenticeshipCommand. 

I'm not a advocate of "all classes must be in separate files", but where there are common patterns across teams, we should follow them. If only for consideration of future devs

benagain: I think we've had this discussion before :)  I like being able to F12 to the command handler directly from the controller.  I almost never want to know what the properties of the DTO are, I want to know what effect using the endpoint has on the system.

Yes, this update file was in the wrong folder.  I'm afraid I don't for one second buy the idea that if the classes were in two separate files in the wrong folder that would have been picked up any more that the single file was.

Pauljgraham (approved):

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/UpdateApprenticeshipCommand.cs
Pauljgraham: Well think this is a discussion we need at a higher level. Dan Beavan might the best person to say what patterns we should and shouldn't use. I've corrected it in my branch and moved the files to the correct folder and removed email

Chazzysb (approved):
Approved

PR #377, opened 2021-05-11
Cs 530 continuation id route

Pauljgraham:


src/SFA.DAS.ApprenticeCommitments.Api/Controllers/ApprenticeshipController.cs
```diff
@@ -24,7 +25,7 @@ public ApprenticeshipController(IMediator mediator, IInternalApiClient<Apprentic
 
         [HttpPost]
         [Route("/apprenticeships")]
-        public async Task<IActionResult> AddApprenticeship(CreateApprenticeshipCommand request)
+        public async Task<IActionResult> AddApprenticeship(ApprenticeshipCreatedCommand request)
```
benagain, line 28: This name sounds strange to me - it is both an event (`ApprenticeshipCreated`) *and* a command.

The Jobs function has received the `ApprenticeshipCreated` event.  Is it not, as a result, now giving APIM a `CreateApprenticeship` command?

src/SFA.DAS.ApprenticeCommitments/Apis/CommitmentsV2InnerApi/ApprenticeshipResponse.cs
```diff
@@ -22,5 +22,6 @@ public class ApprenticeshipResponse
         public bool HasHadDataLockSuccess { get; set; }
         public DateTime? OriginalStartDate { get; set; }
         public long ProviderId { get; set; }
+        public long? ContinuationOfId { get; set; }
```
benagain, line 25: I know it's not ours to change, but that is an unweildy name!

src/SFA.DAS.ApprenticeCommitments/Application/Commands/ApprenticeshipCreated/ApprenticeshipCreatedCommandHandler.cs
```diff
@@ -36,10 +33,52 @@ public class CreateApprenticeshipCommandHandler : IRequestHandler<CreateApprenti
...
+                await CreateApprenticeshipAndSendInvitation(command, trainingProvider, course, apprenticeship);
+            }
+            else
+            {
+                await ChangeApprenticeship(command, trainingProvider, course, apprenticeship);
+            }
+            return Unit.Value;
+        }
+
+        private async Task ChangeApprenticeship(ApprenticeshipCreatedCommand command, TrainingProviderResponse trainingProvider,
```
benagain, line 52: Is there much difference between this `ChangeApprenticeship` and the one that happens because of an `ApprenticeshipUpdatedApprovedEvent`?

I suspect there may be a way to make that clearer in the code. I was already thinking of removing some duplication between the Create and Update handlers - they both do essentially the same thing.

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Features/ContinuationOfApprenticeship.feature
```diff
@@ -0,0 +1,39 @@
...
+	Given the following courses exist
+	| Id   | Title                   | Level |
+	| 9001 | Artificial Intelligence | 1     |
+	| 9002 | Zoology                 | 3     |
+
+Scenario: A continuation apprenticeship is recieved and is valid 
+	When the following apprenticeship is posted
+	| ApprenticeshipId | Email         | Employer Name | Employer Account Legal Entity Id | Training Provider Id | Approved On |
+	| 1                | Test@Test.com | Apple         | 123                              | 1002                 | 2015-04-20  |
+	Then the inner API has received the posted values
```
benagain, line 27: Might it be worth mentioning that it is the "Update" part of the API that is called?

src/SFA.DAS.ApprenticeCommitments.Api/Controllers/ApprenticeshipController.cs
Pauljgraham: Agree, but the previous command CreateApprenticeshipCommand is also incorrect. We are not always creating an apprenticeship. We are actually responding to an approved apprenticeship and either creating a new apprentice (registration) or adding an apprenticeship to an existing apprentice

src/SFA.DAS.ApprenticeCommitments/Application/Commands/ApprenticeshipCreated/ApprenticeshipCreatedCommandHandler.cs
Pauljgraham: No there isn't much difference, but we only know its a continuation apprenticeship in the handler

src/SFA.DAS.ApprenticeCommitments.Api.AcceptanceTests/Features/ContinuationOfApprenticeship.feature
Pauljgraham: agreed, done

src/SFA.DAS.ApprenticeCommitments/Application/Commands/ApprenticeshipCreated/ApprenticeshipCreatedCommandHandler.cs
Pauljgraham: I'm happy to refactor the GetExternalData function into a Extension method

PR #400, opened 2021-06-07
Use email address from Commitments rather than simulated event

benagain:


Pauljgraham (changes requested):
Not all apprenticeship records will have an email on them. So you need to ignore those without an apprenticeship Id

benagain:
> 
> 
> Not all apprenticeship records will have an email on them. So you need to ignore those without an apprenticeship Id

Yes, you're right.  Because we've moved from our simulated event (that we only ever receive for apprenticeships we care about) to the real event (that we will receive for all apprenticeships, even ones that we should ignore).

I've added a check.

PR #409, opened 2021-06-14
Am 449 create contentful service in the apim project

sscaife:


sscaife:
bit premature with this PR, cancelling and reissuing when ready. Have objects in the wrong projects and it doesn't keep in line with other code in the codebase.

PR #410, opened 2021-06-14
Am 449 create contentful service in the apim project & AM 448 Create Get Article Endpoint in APIM using hub and slug

sscaife:


dashton82:
covered in #411

PR #412, opened 2021-06-15
TM-29 List available pledges

ConorTill:
Change base to main prior to merge

See also: 

* https://github.com/SkillsFundingAgency/das-employer-config/pull/913
* https://github.com/SkillsFundingAgency/das-shared-packages/pull/515
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-api/pull/13
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-web/pull/26


src/SFA.DAS.LevyTransferMatching/Interfaces/ILevyTransferMatchingService.cs
```diff
@@ -11,5 +11,6 @@ public interface ILevyTransferMatchingService
         Task<IEnumerable<Tag>> GetSectors();
         Task<IEnumerable<Tag>> GetJobRoles();
         Task<PledgeReference> CreatePledge(Pledge pledge);
+        Task<IEnumerable<Pledge>> GetPledges();
```
chrisfoster76, line 14: I'm not sure that the Pledge class should be used here, given that it is also used in the CreatePledge method too. I don't like the idea of reusing classes. I suspect if we looked there'd be properties that go unused in one or other of the code paths due to the reuse. The classes used by each method should be defined in their own namespaces I'd have thought, be it api requests or api responses, as EI seems to do.

src/SFA.DAS.LevyTransferMatching/Models/Pledge.cs
```diff
@@ -1,13 +1,16 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace SFA.DAS.LevyTransferMatching.Models
 {
-    public class Pledge
+    public class Pledge : PledgeReference
```
chrisfoster76, line 6: As per my previous comment about multiple uses of this class, but also I'm not convinced inheritence is really needed here, just to inherit a couple of properties.

chrisfoster76 (commented):
Some comments about potential smells.

chrisfoster76 (approved):
Approved, but some changes will have to be made subsequently

PR #414, opened 2021-06-18
[TM-35] Option To Apply For Pledge

mmiddleton3301:
Summary:

* Pulling a single pledge from the inner API.
* Pulling back accounts that a user has access to from the `SFA.DAS.EmployerAccounts.Api`.

See also:

* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-api/pull/15
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-web/pull/29
* https://github.com/SkillsFundingAgency/das-employer-config/pull/926
* https://github.com/SkillsFundingAgency/das-employer-config/pull/940
* https://github.com/SkillsFundingAgency/das-employer-config/pull/941


src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -20,13 +21,37 @@ public PledgeController(IMediator mediator)
...
         }
-		
+
+        [HttpGet]
+        [Route("pledges/{pledgeId}")]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        public async Task<IActionResult> GetPledge(int pledgeId)
+        {
+            var result = await _mediator.Send(new GetPledgesQuery()
```
chrisfoster76, line 38: As with the api layer, this should be a separate query with _nothing_ shared

chrisfoster76 (changes requested):

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
mmiddleton3301: This has now been changed, as requested.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -20,13 +22,35 @@ public PledgeController(IMediator mediator)
...
 
             return new OkObjectResult(result.Select(x => (PledgeDto)x));
         }
-		
+
+        [HttpGet]
+        [Route("pledges/{pledgeId}")]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        public async Task<IActionResult> GetPledge(int pledgeId)
```
chrisfoster76, line 37: As discussed on web PR, this should be moved to an Opportunities controller so that the outer api mirrors the web, and we can move to a one-api-call-per-page pattern.

src/SFA.DAS.LevyTransferMatching.Api/Models/AccountDto.cs
```diff
@@ -4,13 +4,15 @@ namespace SFA.DAS.LevyTransferMatching.Api.Models
 {
     public class AccountDto
     {
+        public string EncodedAccountId { get; set; }
         public string DasAccountName { get; set; }
-        public decimal RemainingTransferAllowance { get; set; }
+        public decimal? RemainingTransferAllowance { get; set; }
```
chrisfoster76, line 9: As on the web PR, it's not obvious why this has to change.

src/SFA.DAS.LevyTransferMatching/Application/Queries/GetPledge/GetPledgeQueryHandler.cs
```diff
@@ -0,0 +1,42 @@
+﻿using MediatR;
+using SFA.DAS.LevyTransferMatching.Interfaces;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.LevyTransferMatching.Application.Queries.GetPledge
+{
+    public class GetPledgeQueryHandler : IRequestHandler<GetPledgeQuery, GetPledgeResult>
```
chrisfoster76, line 8: As discussed in web PR, it feels like this should change to be `GetOpportunity`, but we'd still be calling the GetPledge endpoint from the inner api. I guess this is where we map from pledge -> opportunity.

src/SFA.DAS.LevyTransferMatching.Api/Models/AccountDto.cs
mmiddleton3301: Agreed to review DTOs and to discuss request/response convention as part of a team, and to insert tech debt backlog to address this.

See: https://github.com/SkillsFundingAgency/das-levy-transfer-matching-web/pull/29#discussion_r662948503

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
mmiddleton3301: This has been actioned.

src/SFA.DAS.LevyTransferMatching/Application/Queries/GetPledge/GetPledgeQueryHandler.cs
mmiddleton3301: This has been actioned. The service that talks to the inner API still refers to "pledge", but that's because the inner API exposes out "pledges" still, not "opportunities".

src/SFA.DAS.LevyTransferMatching.Api/Models/AccountDto.cs
chrisfoster76: As per comment on web PR, incurring tech debt to deal with some of this is fine, but perhaps not for this Dto, which we'd only have to split/change back later.

chrisfoster76 (approved):
Approved with one suggested change to prevent future reworking

src/SFA.DAS.LevyTransferMatching.Api/Models/AccountDto.cs
mmiddleton3301: I've split this out as requested, into `Account` and `UserAccount` entities, and associated DTOs.

chrisfoster76 (approved):

PR #420, opened 2021-06-25
TM-17 Adding Locations controller to ltm outer api

ConorTill:


src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -31,6 +33,13 @@ public async Task<IActionResult> GetPledges()
         [Route("accounts/{accountId}/pledges")]
         public async Task<IActionResult> CreatePledge(long accountId, [FromBody]CreatePledgeRequest createPledgeRequest)
         {
+            var locationInformation = new List<GetLocationInformationResult>();
```
chrisfoster76, line 36: I think this probably belongs in the (create pledge) handler and not the controller. In the handler you'd be able to inject the location service directly to get the information.

chrisfoster76 (changes requested):
Just one comment around maybe moving a service call into the app layer

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -40,6 +42,7 @@ public async Task<IActionResult> CreatePledge(long accountId, [FromBody]CreatePl
                 JobRoles = createPledgeRequest.JobRoles,
                 Levels = createPledgeRequest.Levels,
                 Sectors = createPledgeRequest.Sectors,
+                Locations = createPledgeRequest.Locations.Select(x => new GetLocationInformationResult { Name = x, GeoPoint = new double[2] }).ToList()
```
chrisfoster76, line 45: I think you're somewhat the victim of some already-tangled code here.

The only class that can be allowed to create `GetLocationInformationResult` must surely be the `GetLocationInformationQueryHandler`. It also seems to add no value by doing this conversion, since the GeoPoint will be an empty array in all cases.

I'd be tempted to leave it as a list of strings on the command itself. I can see why you haven't though:

```
CreatePledgeCommand : Pledge
```

This is wrong. Our commands and queries shouldn't be inheriting from common base classes (especially where those base classes, in turn, inherit other classes...)

We need to break this inheritence, and copy the relevant properties in the command class (the locations can now be a list of strings). 

For the time being, I would then instantiate this `Pledge` class inside the `CreatePledgeCommandHandler` and map all the values across. (I dislike the `Pledge` class full stop, but we'll have to deal with that later). You'll then be able to get the `ILocationLookupService` involved to obtain the GeoPoints.

However, using `GetLocationInformationResult` inside the `Pledge` class also feels wrong. QueryResult classes are meant only to be returned from the application layer, not passed around inside it. Perhaps we should consider location information like this (name & geo point) to be a re-usable type, like the ReferenceData class - see if modelling it like that makes this easier?

chrisfoster76 (changes requested):

chrisfoster76 (dismissed):

chrisfoster76 (approved):

PR #424, opened 2021-07-01
Email user when apprenticeship changes

benagain:


src/SFA.DAS.ApprenticeCommitments.Api/Controllers/ApprenticeshipController.cs
```diff
@@ -39,6 +40,12 @@ public async Task<IActionResult> UpdateApprenticeship(UpdateApprenticeshipComman
             return Accepted();
         }
 
+        [HttpGet("/apprentices/{apprenticeId}")]
+        public async Task<IActionResult> GetApprentice(Guid apprenticeId)
+        {
+            return await new GetRequest<Apprentice>($"/apprentices/{apprenticeId}").Get(_client);
```
Pauljgraham, line 46: This is an unusual pattern, I would have expected _client to be the format to be implemented in the same way as the shared code (Dan put together)

src/SFA.DAS.ApprenticeCommitments.Api/Controllers/GetRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+                };
+            }
+
+            return new ObjectResult(response.Body)
+            {
+                StatusCode = (int)response.StatusCode
+            };
+        }
+    }
+}
```
Pauljgraham, line 31: Does this belong to the shared IInternalApiClient code base. As it may be needed by all

Pauljgraham (approved):

Pauljgraham (approved):

PR #427, opened 2021-07-02
DASD-8885 split outer

adam230594:
- Use new AKS deployment agents
- Pipeline per outer API
- ASP deployment now deploys via shared-infrastructure


deployments/apprentice-commitments.yml
```diff
@@ -0,0 +1,228 @@
...
+  displayName: Deploy to AT
+  variables:
+  - group: DevTest Management Resources
+  - group: AT DevTest Shared Resources
+  - group: AT das-apim-endpoints
+  jobs:
+  - template: ../pipeline-templates/job/arm-deploy.yml
+    parameters:
+      ServiceConnection: SFA-DAS-DevTest-ARM
+      AppRoleAssignmentsServiceConnection: das-app-role-assignments-CDS
```
jack-cloud-platform, line 45: Should the task that uses this service connection only be in this job template, and not deploy-api.yml as well

deployments/apprentice-commitments.yml
```diff
@@ -0,0 +1,228 @@
...
+stages:
+- stage: Deploy_AT
+  dependsOn: []  
+  displayName: Deploy to AT
+  variables:
+  - group: DevTest Management Resources
+  - group: AT DevTest Shared Resources
+  - group: AT das-apim-endpoints
+  jobs:
+  - template: ../pipeline-templates/job/arm-deploy.yml
```
jack-cloud-platform, line 42: Could we pass another parameter for schema file to process? So only the single Outer API's file is processed?

adam230594: Great suggestion. Will get this applied 👍

deployments/apprentice-commitments.yml
adam230594: The app role assignments task is ran in both arm-deploy.yml as well as deploy-api.yml

arm-deploy applies the role assignment for APIM -> outer API's and deploy-api.yaml applies the role assignment for outer API's -> inner API's

pipeline-templates/job/arm-deploy.yml
```diff
@@ -2,12 +2,12 @@ parameters:
   ServiceConnection: ''
   AppRoleAssignmentsServiceConnection: ''
   Environment: ''
+  SchemaFilePath: '*.schema.json'
```
jack-cloud-platform, line 5: Could this have an empty default value?

jack-cloud-platform (approved):

PR #432, opened 2021-07-09
CS-872 - Send registration reminders via messaging

benagain:


src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
```diff
@@ -41,10 +45,14 @@ public class CreateApprenticeshipCommandHandler : IRequestHandler<CreateApprenti
 
             var id = Guid.NewGuid();
 
+            _logger.LogInformation($"Creating registraion for `{apprentice.FirstName} {apprentice.LastName}`");
```
Pauljgraham, line 48: We shouldn't be logging PII data

Pauljgraham (commented):
One comment about PII

src/SFA.DAS.ApprenticeCommitments/Application/Commands/CreateApprenticeship/CreateApprenticeshipCommandHandler.cs
benagain: Removed

Pauljgraham (approved):

PR #439, opened 2021-07-13
Ei 835 clawback payments

David-B-Read:


src/SFA.DAS.EmployerIncentives/Models/ClawbackStatus.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerIncentives.Models
+{
+    public class ClawbackStatus
+    {
+        public decimal? ClawbackAmount { get; set; }
+        public DateTime? ClawbackDate { get; set; }
+    }
```
ruslanss, line 9: in what context are these going to be `null`?
also Clawback prefix is implied and not needed imho 

```suggestion
        public decimal Amount { get; set; }
        public DateTime Date { get; set; }
```

ruslanss: ah the prop name prefix is consistent with PaymentStatus so i guess is ok

ruslanss (approved):

ruslanss (approved):

PR #440, opened 2021-07-13
Tm 22 enter business sector and location

ConorTill:
Must be merged after TM-38 and TM-17


src/SFA.DAS.LevyTransferMatching/Application/Queries/Opportunity/GetSector/GetSectorQueryHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+        public GetSectorQueryHandler(IReferenceDataService referenceDataService, ILocationLookupService locationLookupService, ILevyTransferMatchingService levyTransferMatchingService)
+        {
+            _referenceDataService = referenceDataService;
+            _locationLookupService = locationLookupService;
+            _levyTransferMatchingService = levyTransferMatchingService;
+        }
+
+        public async Task<GetSectorQueryResult> Handle(GetSectorQuery request, CancellationToken cancellationToken)
+        {
+            var locationInformation = await _locationLookupService.GetLocationInformation(request.Postcode, 0, 0, true);
```
chrisfoster76, line 24: It looks like we could await all of these at the same time.

chrisfoster76 (approved):

PR #441, opened 2021-07-14
Tm 97 tech debt

chrisfoster76:


src/SFA.DAS.LevyTransferMatching.Api/Models/Pledges/GetLevelResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using SFA.DAS.LevyTransferMatching.Models.ReferenceData;
+
+namespace SFA.DAS.LevyTransferMatching.Api.Models.Pledges
```
DannyGB, line 4: The reference data items cross over Pledges and Opportunities, should they not be in a namepace of their own rather than `Pledges` or are you advocating here having different GetLevelResponse, GetSectorResponse and GetJobRoleResponse objects for the Opportunities?

DannyGB (approved):

src/SFA.DAS.LevyTransferMatching.Api/Models/Pledges/GetLevelResponse.cs
chrisfoster76: Separate responses for each request; no re-use, no crossover

PR #445, opened 2021-07-17
CS-902-ContentSweep2

Chazzysb:
CS-902-ContentSweep2


Pauljgraham (commented):
Looks fine, but there will be some tests which could be extended to cover the mapping of the duration from the courses API to the call to the inner api

benagain (approved):

Pauljgraham (approved):

PR #450, opened 2021-07-22
Ignore `ApprenticeshipUpdated` events for Framework apprenticeships

benagain:


src/SFA.DAS.ApprenticeCommitments/Application/Commands/UpdateApprenticeship/UpdateApprenticeshipCommandHandler.cs
```diff
@@ -66,21 +60,33 @@ public class UpdateApprenticeshipCommandHandler : IRequestHandler<UpdateApprenti
...
+            {
+                _logger.LogInformation("Apprenticeship {apprenticeshipId} does not have an email, no point in calling Apprentice Commitments", apprenticeship.Id);
+                return default;
+            }
+
+            if (apprenticeship.CourseCode.Contains("-"))
+            {
+                _logger.LogInformation("Apprenticeship {apprenticeshipId} is for a framework, no point in calling Apprentice Commitments", apprenticeship.Id);
+                return default;
+            }
```
Pauljgraham, line 82: We could log this second one as a Warning as it should never happen. It would mean we have put an email against an old apprenticeship record. But I'll leave that up to you.

Pauljgraham (approved):
One tiny suggestion up to you. if you want to

src/SFA.DAS.ApprenticeCommitments/Application/Commands/UpdateApprenticeship/UpdateApprenticeshipCommandHandler.cs
benagain: Good idea, I've done that 👍

Chazzysb (approved):
Approved

PR #462, opened 2021-07-30
SV-925 Approvals Cache

CodeWolf86:
PR to:
- Update the Approvals endpoint to return all versions via the export endpoint
- Expose the new Version Major / Minor fields
- Deal with version now being a string instead of decimal and propagating to other responses (FAT, Approvals / Assessors)
-  I don't believe Fat uses version so hasn't had a further impact though has on assessors
- Also removing options endpoint from assessors


ben1stone (dismissed):

src/SFA.DAS.Approvals.Api/Models/GetStandardResponse.cs
```diff
@@ -2,35 +2,53 @@
...
     {
-        public int Id { get; set; }
+        public string StandardUId { get; set; }
+        public string IfateReferenceNumber { get; set; }
+        public int LarsCode { get; set; }
+        public string Status { get; set; }
         public string Title { get; set; }
-        public int Level { get ; set ; }
+        public int Level { get; set; }
+        public string Version { get; set; }
```
StuartReeve102, line 17: This refers to multiple classes, where you have added a string Version and interger VersionMajor and VersionMinor, properties, shouldn't the string Version property be a calculated property based on VersionMajor and VersionMinor?

StuartReeve102 (changes requested):

src/SFA.DAS.Approvals.Api/Models/GetStandardResponse.cs
CodeWolf86: Given the data from Ifate comes in as a string, it seemed best to store that as is so we hold the original value. The VersionMajor VersionMinor is the calculated aspect from splitting the original field. The sole purpose of the split was purely for ordering purposes at the database level. It is a little bit duplicated but it seemed the simplest option.

StuartReeve102 (approved):

ben1stone (approved):

PR #470, opened 2021-08-11
[TM-47] - View list of applications

DannyGB:
Requires TM-47 LTM Api and Web branches

### DIFF
--- a/src/SFA.DAS.LevyTransferMatching.Api/Authentication/PolicyNames.cs
+++ b/src/SFA.DAS.LevyTransferMatching.Api/Authentication/PolicyNames.cs
@@ -2,6 +2,6 @@
 {
     public static class PolicyNames
     {
-        public const string PledgeAccount = "PledgeAccount";
+        public const string PledgeAccess = "PledgeAccess";
     }
 }
--- a/src/SFA.DAS.LevyTransferMatching.Api/Authentication/PolicyNames.cs
+++ b/src/SFA.DAS.LevyTransferMatching.Api/Authentication/PolicyNames.cs
@@ -1,0 +1,7 @@
+﻿namespace SFA.DAS.LevyTransferMatching.Api.Authentication
+{
+    public static class PolicyNames
+    {
+        public const string PledgeAccount = "PledgeAccount";
+    }
+}
--- a/src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
+++ b/src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using MediatR;
 using Microsoft.AspNetCore.Mvc;
 using SFA.DAS.LevyTransferMatching.Api.Models;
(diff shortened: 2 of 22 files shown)


src/SFA.DAS.LevyTransferMatching.Api/Authentication/PolicyNames.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace SFA.DAS.LevyTransferMatching.Api.Authentication
+{
+    public static class PolicyNames
+    {
+        public const string PledgeAccount = "PledgeAccount";
```
chrisfoster76, line 5: I'd be tempted to rename the policy to something where the intent is more clearly expressed, maybe "ManagePledge" or something like that. Likewise the requirement. "PledgeAccount" is just a bit non-obvious in what it _does_ I think?

Other than that though, looks great.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -162,5 +167,25 @@ public async Task<IActionResult> JobRole()
                 return new StatusCodeResult((int)HttpStatusCode.InternalServerError);
             }
         }
+
+        [Authorize(Policy = PolicyNames.PledgeAccount)]
+        [HttpGet]
+        [Route("accounts/{accountId}/pledges/{pledgeId}/applications")]
+        public async Task<IActionResult> PledgeApplications(int pledgeId)
+        {
+            var applications = await _mediator.Send(new GetApplicationsQuery { PledgeId = pledgeId });
```
chrisfoster76, line 176: The pattern we've used elsewhere has avoided multiple calls to the mediator from the controller in favour of a single call which does everything required and returns a single result. This approach also exposes an inner api type (GetStandardsListItem) nested inside of an outer api response type.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        private readonly ILevyTransferMatchingService _levyTransferMatchingService;
+
+        public GetApplicationsQueryHandler(ILevyTransferMatchingService levyTransferMatchingService)
+        {
+            _levyTransferMatchingService = levyTransferMatchingService;
+        }
+
+        public async Task<GetApplicationsQueryResult> Handle(GetApplicationsQuery request, CancellationToken cancellationToken)
+        {
```
chrisfoster76, line 19: As per previous comment, we could go get the standards here as well (and await them both together) and then assemble a single response.

chrisfoster76 (changes requested):

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
DannyGB: So you would create a second object to hold the Standard data and convert from GetStandardsListItem to it prior to returning back from the query handler?

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
DannyGB: Agreed, I'll implement this

chrisfoster76 (approved):

ConorTill (approved):

PR #475, opened 2021-08-20
[TM-25] Select Account

mmiddleton3301:
Updates the APIM layer to use the most recent pattern, and exposes out accounts for a given user.


sscaife (dismissed):

src/SFA.DAS.LevyTransferMatching.Api/Controllers/OpportunityController.cs
```diff
@@ -272,5 +273,26 @@ public async Task<IActionResult> Detail(int opportunityId)
                 return new StatusCodeResult((int)HttpStatusCode.InternalServerError);
             }
         }
+
+        [HttpGet]
+        [Route("opportunities/{opportunityId}/apply")]
+        [ProducesResponseType((int)HttpStatusCode.OK)]
+        public async Task<IActionResult> OpportunityApply(int opportunityId, string userId)
```
chrisfoster76, line 280: See corresponding points about the naming of this in the UI PR.

chrisfoster76 (changes requested):

src/SFA.DAS.LevyTransferMatching.Api/Controllers/OpportunityController.cs
mmiddleton3301: The corresponding APIM classes have also been renamed.

chrisfoster76 (approved):

PR #485, opened 2021-09-02
TM-151 - Issue with DefaultRequestHeader

dashton82:
Change so that the default request header is no longer used for the version or authorization due to the default headers not being thread safe. The version and auth are now added to the request instead

### DIFF
--- a/src/SFA.DAS.SharedOuterApi/Infrastructure/InternalApiClient.cs
+++ b/src/SFA.DAS.SharedOuterApi/Infrastructure/InternalApiClient.cs
@@ -21,19 +21,13 @@ namespace SFA.DAS.SharedOuterApi.Infrastructure
             _azureClientCredentialHelper = azureClientCredentialHelper;
         }
 
-        protected override async Task AddAuthenticationHeader()
+        protected override async Task AddAuthenticationHeader(HttpRequestMessage request)
         {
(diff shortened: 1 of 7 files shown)


src/SFA.DAS.SharedOuterApi/Infrastructure/InternalApiClient.cs
```diff
@@ -21,19 +21,13 @@ public class InternalApiClient<T> : ApiClient<T>, IInternalApiClient<T> where T
             _azureClientCredentialHelper = azureClientCredentialHelper;
         }
 
-        protected override async Task AddAuthenticationHeader()
+        protected override async Task AddAuthenticationHeader(HttpRequestMessage request)
```
chrisfoster76, line 24: Tedious I know, but the parameter name doesn't match the base class, making this a code smell according to static analysis tools....

chrisfoster76 (dismissed):

src/SFA.DAS.SharedOuterApi/Infrastructure/InternalApiClient.cs
dashton82: 👍🏻 renamed and pushed!

chrisfoster76 (approved):

PR #505, opened 2021-10-04
Tm 53 accept funding for an application

sscaife:


ConorTill (commented):
I missed this before but it looks like we don't have any tests around this stuff.

PR #512, opened 2021-10-12
Tm 48 view matched criteria of applications

ConorTill:


src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
```diff
@@ -20,5 +20,82 @@ public class GetApplicationsQueryHandler : GetApplicationsQueryHandlerBase<GetAp
...
+                application.IsLevelMatch = !pledgeResponse.Levels.Any() || pledgeResponse.Levels.Contains(MapLevelIntToString(application.Standard.Level));
+            }
+
+            return new GetApplicationsQueryResult
+            {
+                Applications = applicationsResponse.Applications.Select(x => (GetApplicationsQueryResultBase.Application)x)
+            };
+        }
+
+        public string MapLevelIntToString(int levelInt)
```
sscaife, line 80: can do this far more efficiently using substring, means that if further levels are ever added you don't have to do anything. If the penultimate value is a string then just grab the last character. If anything it's cleaner. 



Also found this (Level)request.Levels.Cast<int>() never used it but looks even better than substring, the value and the description are also the same number so potentially you can cast to string and do away with the switch statement

sscaife (dismissed):

sscaife (approved):

PR #513, opened 2021-10-12
CS-1090 Potential race condition

benagain:
Change so that the default request header is no longer used for the version or authorization due to the default headers not being thread safe. The version and auth are now added to the request instead.

### DIFF
--- a/deployments/employer-contact-forms.yml
+++ b/deployments/employer-contact-forms.yml
@@ -1,0 +1,148 @@
+trigger: none
+pr: none
+
+variables:
+- group: Release Management Resources
+- group: RELEASE das-apim-endpoints
+- name: OuterApiObject
+  value: $(EmployerContactFormsOuterApiObject)
+
+resources:
+  repositories:
+  - repository: self
+  - repository: das-platform-building-blocks
+    type: github
+    name: SkillsFundingAgency/das-platform-building-blocks
+    ref: refs/tags/0.4.12
+    endpoint: SkillsFundingAgency
+  - repository: das-platform-automation
+    type: github
+    name: SkillsFundingAgency/das-platform-automation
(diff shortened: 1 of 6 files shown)


dashton82:
Why not use the client in `SFA.DAS.SharedOuter.Api` that every other outer API is using? Just seems like there is some duplication occurring here, and a chance for things to come out of sync, as this issue was fixed in the shared client nearly 6 weeks ago.

roysbailey:
> Why not use the client in `SFA.DAS.SharedOuter.Api` that every other outer API is using? Just seems like there is some duplication occurring here, and a chance for things to come out of sync, as this issue was fixed in the shared client nearly 6 weeks ago.

I though it was using SFA.DAS.SharedOuter.Api as we talked about specific changes that needed to be made to it, in order to ensure errors on a GET call were bubbled back to the caller?  And those conversations came from the CMAD team?

benagain:
> 
> 
> Why not use the client in `SFA.DAS.SharedOuter.Api` that every other outer API is using? 

Because the existing shared client doesn't provide the returned `content-type`, and forces the user to define a response class.  We wanted to return whatever was returned to us, including JSON formatted error responses.

dashton82:
So the next person who comes along and see's that same problem - are they going to create their own version to return the HttpResponseMessage? Or extend the GetApiClient to have a new method on it for everyone else to use? Or flipping it on its head if you had found the problem with the DefaultRequestHeader usage would you have just fixed yours and left the one in shared?

It does force a response type to be supplied for the reason that we don't want to tightly couple the inner api through to the outer api consumer through APIM - it defeats the point of it if all we are doing is passing through. You are now in the situation where you cant change your inner api without affecting the consumers of the outer api client.

PR #533, opened 2021-10-29
[TM-54] Decline Funding For An Approved Application & [TM-132] Confirmation Page for Declined Funds (Receiver)

mmiddleton3301:
Includes:

* The necessary outer API changes for declining funding for an approved application.
* APIM changes required for the confirmation page when declining accepted funds.

See also:

* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-api/pull/59
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-web/pull/111
* https://github.com/SkillsFundingAgency/das-data-management/pull/252
* https://github.com/SkillsFundingAgency/das-employer-config/pull/1049
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-functions/pull/10


src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
```diff
@@ -104,5 +105,32 @@ public async Task<IActionResult> DebitApplication(DebitApplicationRequest reques
...
             }
         }
+
+        [Route("application-funding-declined")]
+        [HttpPost]
+        public async Task<IActionResult> ApplicationFundingDeclined(ApplicationFundingDeclinedRequest request)
+        {
+            try
+            {
+                var result = await _mediator.Send(new CreditPledgeCommand
```
chrisfoster76, line 115: I think we need a small amount of logic adding, and this looks like the best place to put it.

If the request.Amount is zero, then we shouldn't call the Credit endpoint - not only would it achieve nothing, there's a validation rule that prevents zero-valued credits. (Amount can be zero since it's the cost in the current f/y, and an April start will cost nothing).

We might have missed this on the Debit use-case, so I'll add a tech debt task to deal with it.

src/SFA.DAS.LevyTransferMatching/Application/Commands/SetApplicationAcceptance/SetApplicationAcceptanceCommandHandler.cs
```diff
@@ -22,11 +23,18 @@ public SetApplicationAcceptanceCommandHandler(ILogger<SetApplicationAcceptanceCo
...
-                throw new NotImplementedException();
-            }
+            HttpStatusCode httpStatusCode;
+
+            if (request.Acceptance == Types.ApplicationAcceptance.Accept)
+                httpStatusCode = await AcceptFunding(request, cancellationToken);
+            else
+                httpStatusCode = await DeclineFunding(request);
 
+            return httpStatusCode == HttpStatusCode.NoContent;
```
chrisfoster76, line 33: I'd have thought checking for any success code would be sufficient, rather than being this specific. Why would we want an error here if a change were made on the endpoint to return content?

I'm also not convinced about returning a bool here vs a specific type vs no return at all, and allowing any exception to bubble up, but it looks like this was part of the original version and not yours, so we can look at it with a tech debt task.

chrisfoster76 (changes requested):
One small but important point about excluding zero-value credits.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
mmiddleton3301: I've added logic in to check the amount - if it's zero or lower, then the inner API isn't called.

src/SFA.DAS.LevyTransferMatching/Application/Commands/SetApplicationAcceptance/SetApplicationAcceptanceCommandHandler.cs
mmiddleton3301: OK, happy to review this as a tech debt task.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
```diff
@@ -104,5 +105,37 @@ public async Task<IActionResult> DebitApplication(DebitApplicationRequest reques
...
+                {
+                    PledgeId = request.PledgeId,
+                    Amount = request.Amount,
+                    ApplicationId = request.ApplicationId
+                });
+
+                if (result == null)
```
chrisfoster76, line 122: I'm not keen on returning null as the result of a command, when there is a result object we could use, as this would be unexpected. Couldn't we just add a new property that we can check here to say that the credit was skipped?

src/SFA.DAS.LevyTransferMatching/Application/Commands/CreditPledge/CreditPledgeCommandHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+            _logger = logger;
+        }
+
+        public async Task<CreditPledgeCommandResult> Handle(CreditPledgeCommand request, CancellationToken cancellationToken)
+        {
+            if (request.Amount <= 0)
+            {
+                _logger.LogInformation($"Amount is 0 for pledge {request.PledgeId} - pledge will *not* be credited");
+
+                return null;
```
chrisfoster76, line 27: See other comment

chrisfoster76 (approved):
Couple of minor points, but pre-approved

src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
mmiddleton3301: Done.

src/SFA.DAS.LevyTransferMatching/Application/Commands/CreditPledge/CreditPledgeCommandHandler.cs
mmiddleton3301: I've included a property that indicates whether the `CreditPledge` endpoint was called.

chrisfoster76 (approved):

PR #538, opened 2021-11-05
TM-46 Modify SetApplicationAcceptance path to include withdrawal and adding…

ConorTill:
… WithdrawApplication to LevyTransferMatchingService


src/SFA.DAS.LevyTransferMatching/Application/Commands/SetApplicationAcceptance/SetApplicationAcceptanceCommandHandler.cs
```diff
@@ -22,26 +22,43 @@ public SetApplicationAcceptanceCommandHandler(ILogger<SetApplicationAcceptanceCo
...
+                    UserId = request.UserId,
+                    UserDisplayName = request.UserDisplayName,
+                    AccountId = request.AccountId,
+                    ApplicationId = request.ApplicationId
+                };
 
-            return result.StatusCode == HttpStatusCode.NoContent;
+                var apiRequest = new WithdrawApplicationRequest(request.ApplicationId, request.AccountId, apiRequestData);
+                await _levyTransferMatchingService.WithdrawApplication(apiRequest, cancellationToken);
+                return true;
```
mmiddleton3301, line 56: Could we not check the status of the inner API request here, and return a bool based on that (similarly to how this is done with acceptance)?

src/SFA.DAS.LevyTransferMatching/Application/Services/LevyTransferMatchingService.cs
```diff
@@ -110,6 +110,11 @@ public async Task<ApiResponse<AcceptFundingRequest>> AcceptFunding(AcceptFunding
             return await _levyTransferMatchingApiClient.PostWithResponseCode<AcceptFundingRequest>(request);
         }
 
+        public async Task WithdrawApplication(WithdrawApplicationRequest request, CancellationToken cancellationToken = default)
+        {
+            await _levyTransferMatchingApiClient.PostWithResponseCode<WithdrawApplicationRequest>(request);
```
mmiddleton3301, line 115: See above, we could return the `ApiResponse<WithdrawApplicationRequest>`, so that it can be used in status checking up the chain.

mmiddleton3301 (commented):
Just a couple of points to keep things consistent on the modified `SetApplicationAcceptanceCommandHandler`.

chrisfoster76 (approved):

sscaife (approved):

PR #541, opened 2021-11-09
Tm 180 export application details csv

sscaife:


src/SFA.DAS.LevyTransferMatching.UnitTests/Application/Queries/Pledges/GetApplicationsForDownload/GetApplicationsForDownloadQueryHandlerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+namespace SFA.DAS.LevyTransferMatching.UnitTests.Application.Queries.Pledges.GetApplicationsForDownload
+{
+    public class GetApplicationsForDownloadQueryHandlerTests
+    {
+        private GetApplicationsForDownloadQueryHandler _handler;
+        private Mock<ILevyTransferMatchingService> _service;
+        private Mock<ICoursesApiClient<CoursesApiConfiguration>> _coursesApiClient;
+        private Mock<IReferenceDataService> _referenceDataService;
+        //private GetApplicationsForDownloadQuery _query;
```
chrisfoster76, line 31: Could we remove the commented lines?

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplicationsForDownload/GetApplicationsForDownloadQueryHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            if (!applications.Applications.Any())
+            {
+                return new GetApplicationsForDownloadQueryResult
+                {
+                    Applications = new List<ApplicationForDownloadModel>()
+                };
+            }
+
+            var standardTasks = RetrieveStandardTasks(applications);
```
chrisfoster76, line 52: Standard info can be retrieved from the application directly now (from inner api)

src/SFA.DAS.SharedOuterApi/Models/Application.cs
```diff
@@ -24,12 +24,16 @@ public class Application
         public Standard Standard { get; set; }
         public bool IsNamePublic { get; set; }
         public string Status { get; set; }
+        public IEnumerable<string> PledgeLocations { get; set; }
```
chrisfoster76, line 27: It feels as though too much is being added to the application object coming back from the inner api - perhaps a separate call to get the pledge would be better?

chrisfoster76 (changes requested):

src/SFA.DAS.LevyTransferMatching.UnitTests/Application/Queries/Pledges/GetApplicationsForDownload/GetApplicationsForDownloadQueryHandlerTests.cs
sscaife: yeah sure

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
```diff
@@ -1,10 +1,14 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
chrisfoster76, line 3: Presumably this is an accident

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
```diff
@@ -37,12 +41,8 @@ public async Task<GetApplicationsQueryResult> Handle(GetApplicationsQuery reques
...
 
                 var applicationResult = (GetApplicationsQueryResult.Application)application;
-
-                applicationResult.IsLocationMatch = !pledgeResponse.Locations.Any() || application.Locations.Any();
-                applicationResult.IsSectorMatch = !pledgeResponse.Sectors.Any() || application.Sectors.Any(x => pledgeResponse.Sectors.Contains(x));
-                applicationResult.IsJobRoleMatch = !pledgeResponse.JobRoles.Any() || roles.Any(r => r.Description == application.StandardRoute);
-                applicationResult.IsLevelMatch = !pledgeResponse.Levels.Any() || pledgeResponse.Levels.Select(x => char.GetNumericValue(x.Last())).Contains(application.StandardLevel);
-
+                
+                AddPledgeDetailsToApplication(applicationResult, pledgeResponse, application, roles);
```
chrisfoster76, line 45: As much as I like the implicit operators, I think this stitching together of these bits is a bit messy (i.e. using the implicit operator to do "most of" the mapping and then doing a bit more elsewhere), though I appreciate the "match" fields were being tacked on like that previously.

Could we just remove the implicit operator here and add a mapping method which takes all the required params so the application result gets instantiated and all properties set in one go?

src/SFA.DAS.LevyTransferMatching.Api/Models/Pledges/GetApplicationsResponse.cs
```diff
@@ -1,5 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
chrisfoster76, line 4: Guessing this is unused

chrisfoster76 (changes requested):
Just one comment about some mapping that could be tidied up, and a couple of unused usings.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplicationsForDownload/GetApplicationsForDownloadQueryHandler.cs
sscaife: removed the file now

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetApplications/GetApplicationsQueryHandler.cs
sscaife: added a factory method that takes the parameters and builds the object

ConorTill (approved):

chrisfoster76 (approved):

PR #544, opened 2021-11-10
FAI-107 retrieve vacancies filter by ale

dashton82:


scottywakefield (approved):
possibly some nesting can be removed in the handler but no biggy

scottywakefield (approved):

PR #566, opened 2021-12-02
TM-199 Add functions endpoints for sending pending application emails

ConorTill:


src/SFA.DAS.LevyTransferMatching/Application/Queries/Functions/GetPendingApplicationEmailDataQueryHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+            var emailDataList = new List<GetPendingApplicationEmailDataQueryResult.EmailData>();
+            foreach (var employerAccount in getAccountsResponse.EmployerAccounts)
+            {
+                List<TeamMember> users = await _accountsService.GetAccountUsers(employerAccount.Id);
+                users = users.Where(x => (x.Role == "Owner" || x.Role == "Transactor") && x.CanReceiveNotifications).ToList();
+
+                if (users.Any())
+                {
+                    var getApplicationsResponse = await _levyTransferMatchingService.GetApplications(new GetApplicationsRequest { AccountId = employerAccount.Id });
+                    var pendingApplications = getApplicationsResponse.Applications.Where(x => x.Status == "Pending");
```
chrisfoster76, line 49: Although we don't have an enum in this layer for this, there is a constants class this could be added to, just to get rid of the magic string (well, not get rid of it as such, but hide it away in a single place) - `PledgeStatus` class in `SFA.DAS.LevyTransferMatching.Models.Constants` - you can just add a `Pending` entry in there.

chrisfoster76 (approved):
One trivial suggestion

chrisfoster76 (dismissed):

chrisfoster76 (approved):

PR #578, opened 2021-12-07
Sh/tm 65 update pledge limit to include other pledges

SHESFA:
Changes in Outer API


src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,26 +1,45 @@
...
+        public GetAmountQueryHandler(IAccountsService accountsService, ILevyTransferMatchingService levyTransferMatchingService)
         {
             _accountsService = accountsService;
+            _levyTransferMatchingService = levyTransferMatchingService;
         }
 
         public async Task<GetAmountQueryResult> Handle(GetAmountQuery request, CancellationToken cancellationToken)
         {
-            var account = await _accountsService.GetAccount(request.EncodedAccountId);
+            // Get the current amount committed to existing pledges first
```
chrisfoster76, line 24: I think the code is sufficiently readable to not warrant a comment.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,26 +1,45 @@
...
+            var response = await _levyTransferMatchingService.GetPledges(new GetPledgesRequest(request.AccountId));
+
+            GetPledgesQueryResult getPledgesQueryResult = new GetPledgesQueryResult
```
chrisfoster76, line 27: Trivial point - we tend to declare using `var` so could we be consistent about this please.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,26 +1,45 @@
...
+                Pledges = response?.Pledges.Select(x => new GetPledgesQueryResult.Pledge
+                {
+                    Amount = x.Amount
+                })
+            };
+
+            int currentTransferAmount = getPledgesQueryResult?.Pledges?.Count() > 0 ? getPledgesQueryResult.Pledges.Sum(p => p.Amount) : 0;
+
+            // Now get the account and get the remaining transfer allowance after subtracting the above already-committed amount
+            var account = await _accountsService.GetAccount(request.AccountId);
```
chrisfoster76, line 38: This call always has to be made and does not depend on the other call - so, we declare both operations together and await them both - they will then execute simultaneously and give us a small performance improvement.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,26 +1,45 @@
...
+
+            GetPledgesQueryResult getPledgesQueryResult = new GetPledgesQueryResult
+            {
+                Pledges = response?.Pledges.Select(x => new GetPledgesQueryResult.Pledge
+                {
+                    Amount = x.Amount
+                })
+            };
+
+            int currentTransferAmount = getPledgesQueryResult?.Pledges?.Count() > 0 ? getPledgesQueryResult.Pledges.Sum(p => p.Amount) : 0;
```
chrisfoster76, line 35: I think this could be rewritten as `getPledgesQueryResult.Pledges.Sum(p => p.Amount);` _if_ you change line 29 so that it does not tolerate `null` responses. Instead, we could check for a null response and return null from this method in this event - then, in the controller, we can check if the query result is null and return a NotFound() if so. (Some of this suggestion does not directly relate to your code change so apologies, but this would be a good time to pick it up).

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -115,11 +115,11 @@ public async Task<IActionResult> CreatePledge(long accountId, [FromBody]CreatePl
 
         [HttpGet]
         [Route("accounts/{accountId}/pledges/create/amount")]
-        public async Task<IActionResult> Amount(string accountId)
+        public async Task<IActionResult> Amount(long accountId)
         {
             try
             {
-                var queryResult = await _mediator.Send(new GetAmountQuery{EncodedAccountId = accountId });
+                var queryResult = await _mediator.Send(new GetAmountQuery{ AccountId = accountId });
```
chrisfoster76, line 122: See previous comment - now would be a good time to check if the queryresult is `null` and, if it is, return a `NotFound` response from this action.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQuery.cs
```diff
@@ -5,5 +5,7 @@ namespace SFA.DAS.LevyTransferMatching.Application.Queries.Pledges.GetAmount
     public class GetAmountQuery : IRequest<GetAmountQueryResult>
     {
         public string EncodedAccountId { get; set; }
+
+        public long AccountId { get; set; }
```
chrisfoster76, line 9: I suspect that this value replaces `EncodedAccountId`, in which case the latter should be removed.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,26 +1,45 @@
...
+
+            // Now get the account and get the remaining transfer allowance after subtracting the above already-committed amount
```
chrisfoster76, line 37: Again, this comment isn't really required as it's pretty readable without it

chrisfoster76 (changes requested):
A few comments, all pretty trivial, nothing major.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
SHESFA: Comment removed

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
SHESFA: Var used now.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
SHESFA: Good point.  Done now.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQuery.cs
SHESFA: You mean AccountId should be removed?  This is still referenced at 6 places?

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
SHESFA: Comment removed

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
SHESFA: Done as suggested.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
SHESFA: Done as suggested.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQuery.cs
chrisfoster76: No, I meant `EncodedAccountId` should be removed, assuming it is no longer referenced?

SHESFA: This needs some further investigation as removing it causes at least one call to fail.  Might need changes in the inner API.  Will look at it and get back asap.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,27 +1,49 @@
...
+
+            var getPledgesQueryResult = new GetPledgesQueryResult
+            {               
+                Pledges = pledges.Result.Pledges.Select(x => new GetPledgesQueryResult.Pledge
+                {
+                    Amount = x.Amount
+                })
+            };
+
+            int currentTransferAmount = getPledgesQueryResult.Pledges.Sum(p => p.Amount);
```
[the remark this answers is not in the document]
SHESFA: Var used now.

src/SFA.DAS.LevyTransferMatching/Application/Queries/Pledges/GetAmount/GetAmountQueryHandler.cs
```diff
@@ -1,27 +1,49 @@
...
         public async Task<GetAmountQueryResult> Handle(GetAmountQuery request, CancellationToken cancellationToken)
         {
-            var account = await _accountsService.GetAccount(request.EncodedAccountId);
+            var pledges = _levyTransferMatchingService.GetPledges(new GetPledgesRequest(request.AccountId));
+
+            var account = _accountsService.GetAccount(request.AccountId);
+
+            await Task.WhenAll(pledges, account);
+
+            if (pledges == null || pledges.Result == null)
```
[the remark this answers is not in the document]
SHESFA: Done

src/SFA.DAS.LevyTransferMatching.UnitTests/Application/Queries/Pledges/GetAmount/GetAmountQueryHandlerTests.cs
```diff
@@ -13,28 +15,43 @@ public class GetAmountQueryHandlerTests
...
 
         [Test]
         public async Task Handle_Result_Has_Correct_RemainingTransferAllowance()
         {
             var result = await _handler.Handle(_query, CancellationToken.None);
-            Assert.AreEqual(result.RemainingTransferAllowance, _account.RemainingTransferAllowance);
+
+            var currentAmount = 0;
+
+            foreach (var pledge in _pledges.Pledges)
```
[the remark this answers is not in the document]
SHESFA: Yes I tried to do that but couldn't get it working.  Working now after correctly importing System.Linq outside resharper.
Change done.

chrisfoster76 (approved):

PR #582, opened 2021-12-14
Cs 1289 apprentice portal apim

Chazzysb:
Cs 1289 apprentice portal apim


src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        [Test, MoqAutoData]
+        public async Task TestGetApprenticeHomepageQuery(            
+            [Frozen] Mock<IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration>> accountsApiClient,
+            [Frozen] Mock<IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration>> commitmentsApiClient,
+            GetApprenticeHomepageQueryHandler sut
+            )
+        {
+            //Arrange
+            var apprenticeId = new Guid("80e8b73c-5c3a-11ec-bf63-0242ac130002");
+            string firstName = "testFirstName", lastName = "testLastName", courseName = "course1", employerName = "employer1";
```
Pauljgraham, line 32: I forgot we could do this!

src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        [Test, MoqAutoData]
+        public async Task TestGetApprenticeHomepageQuery(            
+            [Frozen] Mock<IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration>> accountsApiClient,
+            [Frozen] Mock<IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration>> commitmentsApiClient,
+            GetApprenticeHomepageQueryHandler sut
+            )
+        {
+            //Arrange
+            var apprenticeId = new Guid("80e8b73c-5c3a-11ec-bf63-0242ac130002");
```
Pauljgraham, line 31: I would have thought 

... = Guid.NewGuid();

would be easier, but its the same result.

src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            // Assert
+            Assert.NotNull(result.apprenticeHomepage);
+            Assert.AreEqual(result.apprenticeHomepage.ApprenticeId, apprenticeId);
+            Assert.AreEqual(result.apprenticeHomepage.FirstName, firstName);
+            Assert.AreEqual(result.apprenticeHomepage.LastName, lastName);
+            Assert.AreEqual(result.apprenticeHomepage.CourseName, courseName);
+            Assert.AreEqual(result.apprenticeHomepage.EmployerName, employerName);
+            Assert.IsTrue(result.apprenticeHomepage.ApprenticeshipComplete);
+        }
```
Pauljgraham, line 67: I don't think we can get to the home page, without having an account, but there is a possibility they don't have an apprenticeship. Perhaps we should test for this scenario as well.

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+        public GetApprenticeHomepageQueryHandler(
+            IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration> accountsApiClient,
+            IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration> commitmentsApiClient)
+            => (_accountsApiClient, _commitmentsApiClient) = (accountsApiClient, commitmentsApiClient);
+
+        public async Task<GetApprenticeHomepageQueryResult> Handle(GetApprenticeHomepageQuery request, CancellationToken cancellationToken)
+        {
+            var apprentice = await _accountsApiClient.Get<Apprentice>(new GetApprenticeRequest(request.ApprenticeId));
+            var apprenticeship = await _commitmentsApiClient.Get<GetApprenticeApprenticeshipsResult>(new GetApprenticeApprenticeshipsRequest(request.ApprenticeId));
+
```
Pauljgraham, line 26: We could do this in parallel 

var apprenticeTask = _accountsApiClient.Get....
var apprenticeshipTask = _commitmentsApiClient.Get....

Task.WhenAll(apprenticeTask, apprenticeshipTask);

var apprentice = await apprenticeTask;
var apprenticeship = await apprenticeshipsTask;

As the calls are independent of each other.

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.ApprenticePortal.Application.Homepage.Queries
+{
+    public class GetApprenticeHomepageQueryHandler : IRequestHandler<GetApprenticeHomepageQuery, GetApprenticeHomepageQueryResult>
+    {
+        private readonly IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration> _accountsApiClient;
+        private readonly IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration> _commitmentsApiClient;
+
```
Pauljgraham, line 16: would this be better named _apprenticeCommitmentsApiClient or _apprenticeshipsApiClient. The name implies to me that we are calling the approvals commitments api. But it's up to you

src/SFA.DAS.ApprenticePortal/Infrastructure/ApprenticeAccountsHealthCheck.cs
```diff
@@ -0,0 +1,43 @@
...
+            if ((int)response == 200)
+            {
+                return HealthCheckResult.Healthy(HealthCheckResultDescription, data);
+            }
+
+            _logger.LogWarning($"{HealthCheckResultDescription} API ping failed : [Code: {response}]");            
+             return HealthCheckResult.Unhealthy(HealthCheckResultDescription, null, data);
+        }
+    }
+}
```
Pauljgraham, line 43: Interestingly I have added my version of this class to the shared infrastructure codebase.

src/SFA.DAS.ApprenticePortal/Infrastructure/ApprenticeCommitmentsHealthCheck.cs
```diff
@@ -0,0 +1,43 @@
...
+            if ((int)response == 200)
+            {
+                return HealthCheckResult.Healthy(HealthCheckResultDescription, data);
+            }
+
+            _logger.LogWarning($"{HealthCheckResultDescription} API ping failed : [Code: {response}]");
+            return HealthCheckResult.Unhealthy(HealthCheckResultDescription, null, data);
+        }
+    }
+}
```
Pauljgraham, line 43: This class already exists in the Shared Infrastructure code base.

src/SFA.DAS.ApprenticePortal/Models/ApprenticeHomepage.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace SFA.DAS.ApprenticePortal.Models
+{
+    public class ApprenticeHomepage
+    {
+        public Guid ApprenticeId { get; set; }
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public string CourseName { get; set; }
+        public string EmployerName { get; set; }
+        public bool? ApprenticeshipComplete { get; set; }
```
Pauljgraham, line 12: I think we will also need the Apprenticeship Stopped Status on the home page

src/SFA.DAS.ApprenticePortal/Models/Apprenticeship.cs
```diff
@@ -0,0 +1,32 @@
...
+        public int CourseDuration { get; set; }
+        public DateTime PlannedStartDate { get; set; }
+        public DateTime PlannedEndDate { get; set; }
+        public DateTime ConfirmBefore { get; set; }
+        public DateTime? ConfirmedOn { get; set; }
+        public long RevisionId { get; set; }
+        public ChangeOfCircumstanceNotifications ChangeOfCircumstanceNotifications { get; set; }
+        public DateTime ApprovedOn { get; set; }
+        public DateTime? LastViewed { get; set; }
+    }
```
Pauljgraham, line 31: We appear to be returning more details here than we need. If we are not using fields there is a very good argument to say they should be present. It becomes a little bit of a maintenance problem. Ie. if ChangeOfCircumstanceNotifications is updated do we have to update it here (if our consuming web site has no interest in it)?

src/SFA.DAS.SharedOuterApi/Infrastructure/HealthCheck/ApprenticeAccountsApiHealthCheck.cs
```diff
@@ -0,0 +1,49 @@
...
+using SFA.DAS.SharedOuterApi.InnerApi.Requests;
+using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.SharedOuterApi.Infrastructure.HealthCheck
+{
+    public class ApprenticeAccountsApiHealthCheck : IHealthCheck
+    {
+        private const string HealthCheckResultDescription = "Apprentice Accounts Api check";
+
+        private readonly IApprenticeCommitmentsApiClient<ApprenticeAccountsApiConfiguration> _apiClient;
```
Pauljgraham, line 18: shouldn't this be IApprenticeAccountsApiClient<>

src/SFA.DAS.SharedOuterApi/Infrastructure/HealthCheck/ApprenticeAccountsApiHealthCheck.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                return HealthCheckResult.Healthy(HealthCheckResultDescription,
+                    new Dictionary<string, object> { { "Duration", durationString } });
+            }
+
+            _logger.LogWarning($"ApprenticeAccounts API ping failed : [Code: {response}]");
+            return HealthCheckResult.Unhealthy(HealthCheckResultDescription);
+        }
+    }
+}
```
Pauljgraham, line 49: I'm a little confused about this endpoint, this healthcheck also exists in the apprenticeportal infrastructure site, so I suspect this one isn't being used (probably why the issue above isn't causing problems). 

Just to add to the confusion I have also added this heathcheck here in the approvals APIM rework. But we can take whichever arrives first in master

src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
Chazzysb: changed to NewGuid()

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
Chazzysb: Added helper class now tasks are called in parallel

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
Chazzysb: commitmentsApi is embedded in my head!

src/SFA.DAS.ApprenticePortal/Models/ApprenticeHomepage.cs
Chazzysb: Added necessary columns required on the front end

src/SFA.DAS.ApprenticePortal/Infrastructure/ApprenticeCommitmentsHealthCheck.cs
Chazzysb: I think this is left over from when I was copying the commitments project - now removed as the one in shared is actually used

src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
Chazzysb: test added

src/SFA.DAS.ApprenticePortal/Models/Apprenticeship.cs
Chazzysb: reduced dtos to required columns only

src/SFA.DAS.SharedOuterApi/Infrastructure/HealthCheck/ApprenticeAccountsApiHealthCheck.cs
Chazzysb: good spot!

src/SFA.DAS.SharedOuterApi/Infrastructure/HealthCheck/ApprenticeAccountsApiHealthCheck.cs
Chazzysb: yes the one in the portal is there as I was copying the commitments setup - ive removed it from the portal as its not used

src/SFA.DAS.ApprenticePortal/Application/Extensions/TaskEx.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public static class TaskEx
+    {
+        public static async Task<(T1, T2)> AwaitAll<T1, T2>(Task<T1> t1, Task<T2> t2)
+        {
+            await Task.WhenAll(t1, t2);
+            return (t1.Result, t2.Result);
+        }
+    }
+}
```
Pauljgraham, line 13: My only 2 reservations against this pattern is 

1. it's actually quite hard to debug, if one of the tasks is failing. You have to refactor the code to determine which task is failing.
2. What happens when 3 or more tasks can be executed at the same time, be better as an input array, but that's a minor point

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+            IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration> accountsApiClient,
+            IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration> commitmentsApiClient)
+            => (_accountsApiClient, _commitmentsApiClient) = (accountsApiClient, commitmentsApiClient);
+
+        public async Task<GetApprenticeHomepageQueryResult> Handle(GetApprenticeHomepageQuery request, CancellationToken cancellationToken)
+        {
+            var (_apprentice, _apprenticeships) = await TaskEx.AwaitAll(
+                Task.FromResult(_accountsApiClient.Get<Apprentice>(new GetApprenticeRequest(request.ApprenticeId))).Result,
+                Task.FromResult(_commitmentsApiClient.Get<GetApprenticeApprenticeshipsResult>(new GetApprenticeApprenticeshipsRequest(request.ApprenticeId))).Result
+                );
```
Pauljgraham, line 29: This doesn't quite look right to me, The layout I was expecting was as follows

var (_apprentice, _apprenticeships) = await TaskEx.AwaitAll(
                _accountsApiClient.Get<Apprentice>(new GetApprenticeRequest(request.ApprenticeId)),
                _commitmentsApiClient.Get<GetApprenticeApprenticeshipsResult>(new GetApprenticeApprenticeshipsRequest(request.ApprenticeId)).Result
                );
                
 This code looks like it's evaluating the task prior to passing the result to the TaskEx.AwaitAll function.

src/SFA.DAS.ApprenticePortal/Models/RolesAndResponsibilitiesConfirmations.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace SFA.DAS.ApprenticePortal.Models
+{
+    [Flags]
+    public enum RolesAndResponsibilitiesConfirmations : byte
+    {
+        ApprenticeRolesAndResponsibilitiesConfirmed = 1,
+        EmployerRolesAndResponsibilitiesConfirmed = 2,
+        ProviderRolesAndResponsibilitiesConfirmed = 4
+    }
+}
```
Pauljgraham, line 12: think it can be removed

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            return new GetApprenticeHomepageQueryResult
+            {
+                apprenticeHomepage = new ApprenticeHomepage
+                {                    
+                    apprentice = _apprentice,
+                    apprenticeship = _apprenticeships?.apprenticeships.FirstOrDefault()
+                },                
+            };
```
Pauljgraham, line 38: I might be wrong here, but I think if the apprentice personal details do not exist in the ApprenticeAccounts API, It will return a 404 response and this will result in a null object. Therefore we will lose the ApprenticeId if we model the ApprenticeHomePage like this.

src/SFA.DAS.ApprenticePortal.UnitTests/Application/Homepage/Queries/GetApprenticeHomepageQueryTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                    .With(p => p.ConfirmedOn, DateTime.Now)
+                    .Create()};
+        }
+
+        private void SetupBlankMoqApprenticeships()
+        {
+            _moqApprenticeships = new List<Apprenticeship>();                
+        }
+    }
+}
```
Pauljgraham, line 116: A matter worth discussing, I missed it earlier. Is it better to test this code using Wiremock or using this traditional method.   

I'm not totally convinced client.Get<> is working the way we think it is and wiremock to me seems a much better testing strategy, as it's really testing the boundaries of the system and not making assumptions about the internal components. Ben may have an opinion on this.

 Wiremock also allows us to update the internals without having to refactor our tests

src/SFA.DAS.ApprenticePortal/Models/ApprenticeHomepage.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace SFA.DAS.ApprenticePortal.Models
+{
+    public class ApprenticeHomepage
+    {
+        public Apprentice apprentice { get; set; }
+        public Apprenticeship apprenticeship { get; set; }
+    }
```
Pauljgraham, line 9: I definitely think the names should be have the first letter as a capital. 

The model might more sense as

public class ApprenticeHomepage
    {
        public ApprenticeId { get; set;
        public ApprenticePersonalDetails ApprenticePersonalDetails { get; set; }
        public Apprenticeship LatestApprenticeship { get; set; }
    }

Pauljgraham (commented):
Hi Chase, just a few more comments. Also we should discuss the value of implementing tests using WireMock. The way we are doing here is a little older in style and it leads to us making assumptions about internal components. Ben may have an opinion on this

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
benagain: Yes, I don't think those `Task.FromResult` calls are needed.  They were in the example just to turn an `int` into a `Task<int>`, but the `Get` functions already return `Task`s.

I don't think there's any need for `.Result` in this calling code.
```
            var (_apprentice, _apprenticeships) = await TaskEx.AwaitAll(
                _accountsApiClient.Get<Apprentice>(new GetApprenticeRequest(request.ApprenticeId)),
                _commitmentsApiClient.Get<GetApprenticeApprenticeshipsResult>(new GetApprenticeApprenticeshipsRequest(request.ApprenticeId))
                );
```

src/SFA.DAS.ApprenticePortal/Application/Extensions/TaskEx.cs
Chazzysb: you would need to put a break point in the inner api calls or try catch to debug.

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
Chazzysb: Updated as discussed

src/SFA.DAS.ApprenticePortal/Models/RolesAndResponsibilitiesConfirmations.cs
Chazzysb: thought id already removed it - removed now

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
Chazzysb: the apprentice id is the parameter sent in the querystring, the caller will know.  the api will not 404 but return a null object for the apprentice and apprenticeship in this instance

src/SFA.DAS.ApprenticePortal/Models/ApprenticeHomepage.cs
Chazzysb: its been given the same name as the object it represents - unnecessarily changing names just adds confusion

src/SFA.DAS.ApprenticePortal/Application/Homepage/Queries/GetApprenticeHomepageQueryHandler.cs
Chazzysb: Ive updated the controller - in the event that an accounts apprentice is null then it will now return a 404 NotFound response

Pauljgraham (dismissed):

benagain (approved):

benagain (approved):

PR #611, opened 2022-01-21
TM-52 - Notify the receiving employer of application approval - APIM

SHESFA:


src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
```diff
@@ -52,6 +56,30 @@ public async Task<IActionResult> ApplicationApproved(ApplicationApprovedRequest
             }
         }
 
+        [Route("application-approved-receiver-notification")]
+        [HttpPost]
+        public async Task<IActionResult> ApplicationApprovedReceiverNotification(ApplicationApprovedReceiverNotificationRequest request)
+        {
+            try
+            {
+                var result = await _mediator.Send(new ReceiverApplicationApprovedEmailCommand
```
chrisfoster76, line 65: The `result` variable is not used so need not be assigned.

chrisfoster76 (approved):
One trivial comment

src/SFA.DAS.LevyTransferMatching.Api/Controllers/FunctionsController.cs
SHESFA: var result removed now.

ConorTill (approved):

PR #630, opened 2022-02-09
TM-255 bulk reject applications non javascript version - APIM

jawwadbaig:
TM-255 bulk reject applications non javascript version - APIM


src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -144,6 +145,22 @@ public async Task<IActionResult> ClosePledge(int pledgeId, [FromBody] SetClosePl
             }
         }
 
+        [HttpPost]
+        [Route("accounts/{accountId}/pledges/{pledgeId}/rejectapplications")]
```
chrisfoster76, line 149: Nit-picky, but we should hyphenate the route for readability.

src/SFA.DAS.LevyTransferMatching.Api/Models/ApplicationRejectRequest.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Newtonsoft.Json;
+using System.Collections.Generic;
+
+namespace SFA.DAS.LevyTransferMatching.Api.Models
+{
+    public class ApplicationRejectRequest
+    {
+        public List<string> ApplicationsToReject { get; set; }
```
chrisfoster76, line 8: Why is this a list of strings? I'd have thought a list of integers would be sufficient

src/SFA.DAS.LevyTransferMatching/Application/Commands/RejectApplication/RejectApplicationsCommand.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace SFA.DAS.LevyTransferMatching.Application.Commands.RejectApplication
+{
+    public class RejectApplicationsCommand : IRequest
+    { 
+        public int PledgeId { get; set; }
+        public int AccountId { get; set; }     
+        public string UserId { get; set; }
+        public string UserDisplayName { get; set; }
+        public List<string> ApplicationsToReject { get; set; }
```
chrisfoster76, line 12: As before, I'd have thought a list of integers would be appropriate.

src/SFA.DAS.LevyTransferMatching.Api/Models/ApplicationRejectRequest.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Newtonsoft.Json;
+using System.Collections.Generic;
+
+namespace SFA.DAS.LevyTransferMatching.Api.Models
+{
+    public class ApplicationRejectRequest
```
chrisfoster76, line 6: Why not `RejectApplicationsRequest`?

src/SFA.DAS.LevyTransferMatching/Application/Commands/RejectApplication/RejectApplicationsHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+using SFA.DAS.SharedOuterApi.Interfaces;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using SFA.DAS.LevyTransferMatching.InnerApi.LevyTransferMatching.Requests;
+
+namespace SFA.DAS.LevyTransferMatching.Application.Commands.RejectApplication
+{
+    public class RejectApplicationsHandler : IRequestHandler<RejectApplicationsCommand>
```
chrisfoster76, line 13: The convention (perhaps not 100% adhered to) would make this `RejectApplications**Command**Handler`

chrisfoster76 (changes requested):

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
jawwadbaig: I have updated the route now

src/SFA.DAS.LevyTransferMatching.Api/Models/ApplicationRejectRequest.cs
jawwadbaig: I have refactored it and changed it to list in integers.

src/SFA.DAS.LevyTransferMatching/Application/Commands/RejectApplication/RejectApplicationsCommand.cs
jawwadbaig: I have refactored it and changed it to a list of integers. It makes sense.

src/SFA.DAS.LevyTransferMatching/Application/Commands/RejectApplication/RejectApplicationsHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            var apiRequestData = new RejectApplicationRequestData
+            {
+                UserId = request.UserId,
+                UserDisplayName = request.UserDisplayName
+            };
+            
+            foreach (var applicationId in request.ApplicationsToReject)
+            {
+                var id = int.Parse(applicationId.Split("-")[0]);
```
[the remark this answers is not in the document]
jawwadbaig: Totally makes sense. I was thinking the same way when I looked at that logic. I have refactored it now. We are just passing the application ID as an integer now.

src/SFA.DAS.LevyTransferMatching.Api/Models/ApplicationRejectRequest.cs
jawwadbaig: Yes sure. I have refactored it and renames it now.

src/SFA.DAS.LevyTransferMatching/Application/Commands/RejectApplication/RejectApplicationsHandler.cs
jawwadbaig: I have renames it now.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -247,7 +265,20 @@ public async Task<IActionResult> PledgeApplications(int pledgeId)
                 PledgeStatus = queryResult?.PledgeStatus
             });
         }
-        
+
+        [Authorize(Policy = PolicyNames.PledgeAccess)]
+        [HttpGet]
+        [Route("accounts/{accountId}/pledges/{pledgeId}/applications-account-names")]
```
chrisfoster76, line 271: The pattern is for endpoints in the outer api to have matching routes with the page in the UI which they serve, except that values are not encoded.

Since this page is there to support the `.../applications/reject-applications` `GET` endpoint in the UI, I'd expect it to also have a route of `.../applications/reject-applications`, and for the method, query, handler etc. names to reflect that.

This pattern makes it possible to see which endpoints serve which pages really quickly, and to discourage inappropriate re-use of outer api endpoints.

chrisfoster76 (changes requested):
Just a route change

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
jawwadbaig: Yes sure. I can change that. The only reason I created this with that name is so that this endpoint only returns a list of the DAS Account Names. I will change it to reject-applications (GET) and will also change the related methods to reflect this change.

jawwadbaig: I have changed the GET endpoint and method, query, handler as well.

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
```diff
@@ -144,6 +146,22 @@ public async Task<IActionResult> ClosePledge(int pledgeId, [FromBody] SetClosePl
             }
         }
 
+        [HttpPost]
+        [Route("accounts/{accountId}/pledges/{pledgeId}/reject-applications")]
+        public async Task<IActionResult> RejectApplications(int accountId, int pledgeId, [FromBody] RejectApplicationsRequest request)
```
chrisfoster76, line 151: Minor point, but methods with the same name and access level should be next to one another for clarity.

chrisfoster76 (approved):
Approving - with one request that a method be moved to be next to another overload

src/SFA.DAS.LevyTransferMatching.Api/Controllers/PledgeController.cs
jawwadbaig: Methos for RejectApplications with the same names are moved together now.

chrisfoster76 (dismissed):

chrisfoster76 (approved):

PR #642, opened 2022-02-23
tm 31 filter pledges

hasanrq:


src/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPledgesRequest.cs
```diff
@@ -1,17 +1,28 @@
 ﻿using SFA.DAS.SharedOuterApi.Interfaces;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace SFA.DAS.SharedOuterApi.InnerApi.Requests
 {
     public class GetPledgesRequest : IGetApiRequest
     {
-        public GetPledgesRequest(long? accountId = null)
+        public GetPledgesRequest(IEnumerable<string> sectors = null, long? accountId = null)
```
chrisfoster76, line 9: I think these changes would benefit from refactoring - it's hard to read as it is.

I'd also be tempted to leave accountId as the first param, only because it's more commonly used and more important (and was here first).

chrisfoster76 (changes requested):

src/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPledgesRequest.cs
hasanrq: I have added similar NameValueCollection helper class here and formed querystring thru that.

chrisfoster76 (approved):

src/SFA.DAS.SharedOuterApi/Extensions/NameValueCollectionExtensions.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Web;
+
+namespace SFA.DAS.SharedOuterApi.Extensions
+{
+    public static class NameValueCollectionExtensions
+    {
+        public static string ToQueryString(this NameValueCollection source)
```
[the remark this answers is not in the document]
hasanrq: Unit test added

chrisfoster76 (approved):

PR #652, opened 2022-03-08
QF-294 get feedback attributes outer api

SarahDuncan:
Outer API to surface provider attributes.

### DIFF
--- a/src/SFA.DAS.ApprenticeFeedback.Api.UnitTests/Controllers/WhenGettingProviderAttributes.cs
+++ b/src/SFA.DAS.ApprenticeFeedback.Api.UnitTests/Controllers/WhenGettingProviderAttributes.cs
@@ -1,0 +1,50 @@
+﻿using AutoFixture.NUnit3;
+using FluentAssertions;
+using MediatR;
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+using NUnit.Framework;
+using SFA.DAS.ApprenticeFeedback.Api.Controllers;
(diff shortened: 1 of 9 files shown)


src/SFA.DAS.ApprenticeCommitments.Api/appsettings.json
```diff
@@ -12,4 +12,4 @@
   "Environment": "DEV",
   "Version": "1.0",
   "APPINSIGHTS_INSTRUMENTATIONKEY": ""
-}
```
CodeWolf86: Can we lose this change? we shouldn't really be making changes to other teams projects within our own PR's

src/SFA.DAS.ApprenticeFeedback.Api/AppStart/AddServiceRegistrationExtension.cs
```diff
@@ -17,7 +17,7 @@ public static void AddServiceRegistration(this IServiceCollection services)
             services.AddTransient(typeof(IInternalApiClient<>), typeof(InternalApiClient<>));
 
             services.AddTransient<IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration>, ApprenticeAccountsApiClient>();
-            services.AddTransient<IApprenticeCommitmentsApiClient<ApprenticeCommitmentsApiConfiguration>, ApprenticeCommitmentsApiClient>();
```
CodeWolf86: Should this be commented out?

CodeWolf86 (approved):

PR #654, opened 2022-03-09
Ei 1686 new outer endpoint for employment check

MichaelYoung1981:


src/SFA.DAS.EmployerIncentives/Application/Commands/RegisterEmploymentCheck/RegisterEmploymentCheckCommandHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class RegisterEmploymentCheckCommandHandler : IRequestHandler<RegisterEmploymentCheckCommand, RegisterEmploymentCheckResponse>
+    {
+        private readonly IInternalApiClient<EmploymentCheckConfiguration> _client;
+
+        public RegisterEmploymentCheckCommandHandler(IInternalApiClient<EmploymentCheckConfiguration> client)
+        {
+            _client = client;
+        }
+
+        public async Task<RegisterEmploymentCheckResponse> Handle(RegisterEmploymentCheckCommand command, CancellationToken cancellationToken)
```
David-B-Read, line 21: I thought the established APIM pattern was to call a service from the command handler, rather than the API client directly?

David-B-Read (approved):

PR #655, opened 2022-03-10
PA-223 Call roatp api to get deliverymodel (This may be a new API)

Pauljgraham:


src/SFA.DAS.Approvals.Api/Controllers/ProvidersController.cs
```diff
@@ -43,5 +43,21 @@ public async Task<IActionResult> GetAll()
...
+        {
+            try
+            {
+                var result = await _mediator.Send(new GetProviderCoursesDeliveryModelQuery(providerId, trainingCode));
+                return Ok(result);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error getting Provider Courses Delivery Models for Provider {providerId} and course {trainingCode}", providerId, trainingCode);
+                return BadRequest();
```
benagain, line 59: This seems more like an `InternalServerError` than a bad request don't you think?

A bad request would be if the provider ID was less than 1, or the training code was blank.

benagain (dismissed):
Looks good.  I think maybe we should change the HTTP Status returned when there's an error.

benagain (approved):

benagain (approved):

benagain (approved):

PR #656, opened 2022-03-11
Qf 287 employer feedback ratings api changes

anilkdhiman6:


CodeWolf86 (changes requested):

src/SFA.DAS.FindApprenticeshipTraining.Api/Models/GetProviderFeedbackResponse.cs
```diff
@@ -84,4 +79,20 @@ public enum FeedbackRatingType
         Good = 3,
         Excellent = 4
     }
+
+    public class StrengthDetail
```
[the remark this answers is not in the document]
anilkdhiman6: Updated to :

`    public class FeedbackAttributeDetail
    {
        public string AttributeName { get; set; }
        public int StrengthCount { get; set; }
        public int WeaknessCount { get; set; }

    }
`

src/SFA.DAS.FindApprenticeshipTraining.Api.UnitTests/Models/WhenBuildingGetProviderFeedbackAttributes.cs
```diff
@@ -69,9 +68,9 @@ public void Then_No_Strengths_Returns_Empty_List(GetProvidersListItem source, st
             
             var response = new GetTrainingCourseProviderListItem().Map(source, sectorSubjectArea,1,new List<DeliveryModeType>(),new List<FeedbackRatingType>(), true);
 
-            response.Feedback.FeedbackAttributes.Strengths.Should().BeEmpty();
-            response.Feedback.FeedbackAttributes.Weaknesses.Should().Contain(source.FeedbackAttributes.Select(c => c.AttributeName).ToList());
-            response.Feedback.FeedbackAttributes.Weaknesses.Should().NotBeEmpty();
+            response.Feedback.FeedbackAttributes.FeedbackAttributeDetail.Should().BeEmpty();
+            response.Feedback.FeedbackAttributes.FeedbackAttributeDetail.Should().Contain(source.FeedbackAttributes.Select(c => c.AttributeName).ToList());
+            response.Feedback.FeedbackAttributes.FeedbackAttributeDetail.Should().NotBeEmpty();
```
CodeWolf86, line 73: These tests need to be looked at in more detail. The previous iterations of the unit tests relied on the fact there was two collections, a strength and a weakness. Now there is only one and the assertions completely contradict each other so assume this is a renaming gone wrong. But the build is failing due to these unit tests



` response.Feedback.FeedbackAttributes.FeedbackAttributeDetail.Should().BeEmpty();
 response.Feedback.FeedbackAttributes.FeedbackAttributeDetail.Should().NotBeEmpty();`

CodeWolf86 (changes requested):

src/SFA.DAS.FindApprenticeshipTraining.Api.UnitTests/Models/WhenBuildingGetProviderFeedbackAttributes.cs
anilkdhiman6: Updated tests & renamed according to changes made

src/SFA.DAS.FindApprenticeshipTraining.Api/Models/GetProviderFeedbackResponse.cs
```diff
@@ -53,25 +53,19 @@ public class GetProviderFeedbackAttributeItem
 
     public class GetProviderFeedbackAttributes
     {
-        public List<string> Strengths { get; private set; }
-        public List<string> Weaknesses { get; private set; }
+        public List<FeedbackAttributeDetail> FeedbackAttributeDetail { get; private set; }
 
         public GetProviderFeedbackAttributes Build(List<GetProviderFeedbackAttributeItem> feedbackAttributeItems)
         {
             return new GetProviderFeedbackAttributes
```
CodeWolf86, line 60: Did we want to simplify this object so it's not nested anymore? as rather than it being two properties, it's just one so the nesting seems overkill? Could just be a list of Feedback Attribute Details? There may even be an object that is suitable already mentioned previously as I think that's what it converts from in the first place if you have a look?

src/SFA.DAS.FindApprenticeshipTraining.Api.UnitTests/Models/WhenBuildingGetProviderFeedbackAttributes.cs
```diff
@@ -133,105 +132,91 @@ public void Then_If_Same_Feedback_Attribute_Score_Then_The_Attribute_With_Most_R
...
+        public void Then_Returns_All_Available_Feedback_Attribute_Strengths_And_Weaknesses(GetProvidersListItem source, string sectorSubjectArea)
         {
             source.FeedbackAttributes = new List<GetFeedbackAttributeItem>
             {
                 new GetFeedbackAttributeItem
                 {
                     AttributeName = "First Attribute",
-                    Strength = 11,
-                    Weakness = 10
+                    Strength = 1,
```
CodeWolf86, line 146: I get what you've done here in terms of equalising the strengths / weaknesses, and it returns all of them.

I don't know if we enforce ordering on the items returned from the API? IE ordered by Total votes ( strength / weakness ) then by strengths / then by weaknesses, which would then allow this test to prove something, as currently it just returns the list with the same values, so it feels like the test itself has become somewhat invalidated.

CodeWolf86 (changes requested):

src/SFA.DAS.FindApprenticeshipTraining.Api/Models/GetProviderFeedbackResponse.cs
anilkdhiman6: Updated

src/SFA.DAS.FindApprenticeshipTraining.Api.UnitTests/Models/WhenBuildingGetProviderFeedbackAttributes.cs
anilkdhiman6: Updated test

CodeWolf86 (approved):

PR #707, opened 2022-05-06
CSP-17 Get standards from courses api

MarkFCain:


src/SFA.DAS.Roatp.CourseManagement/InnerApi/Requests/GetStandardRequest.cs
```diff
@@ -0,0 +1,15 @@
+﻿using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.Roatp.CourseManagement.InnerApi.Requests
+{
+    public class GetStandardRequest : IGetApiRequest
+    {
+        public string GetUrl => $"/api/courses/Standards/{StandardCode}";
```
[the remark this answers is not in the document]
MarkFCain: good spot, I checked against other apim repos (eg Commitments etc) and it doesn't have it, will take off the prefix

src/SFA.DAS.Roatp.CourseManagement.Api/Startup.cs
```diff
@@ -45,6 +46,7 @@ public void ConfigureServices(IServiceCollection services)
 
                 services.AddAuthentication(azureAdConfiguration, policies);
             }
+            services.AddMediatR(GetType().Assembly, typeof(GetProviderCourseQueryHandler).Assembly);
```
DevenKShah, line 49: There is no need to do this.

src/SFA.DAS.Roatp.CourseManagement/Application/Standards/Queries/GetProviderCourseQuery/GetProviderCourseQueryHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+            _courseManagementApiClient = courseManagementApiClient;
+            _logger = logger;
+        }
+
+        public async Task<GetProviderCourseResult> Handle(GetProviderCourseQuery request, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Get Provider Course request received for ukprn {ukprn}, LarsCode {larsCode}", request.Ukprn, request.LarsCode);
+
+            GetStandardResponse standard;
+            try
```
DevenKShah, line 31: Remove the higher level try catch, this should be handled by a middleware

src/SFA.DAS.Roatp.CourseManagement/Application/Standards/Queries/GetProviderCourseQuery/GetProviderCourseQueryHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+                    return null;
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"Error occurred trying to retrieve standard for LarsCode {request.LarsCode}");
+                throw;
+            }
+
+            try
```
DevenKShah, line 46: Remove higher level try catch

src/SFA.DAS.Roatp.CourseManagement/Application/Standards/Queries/GetProviderCourseQuery/GetProviderCourseQueryHandler.cs
MarkFCain: I thought we'd established the middleware in apim doesn't do this, so we have to leave this in.  I'll dive in and have a look

PR #717, opened 2022-05-16
CSP-18-outer-rebase-17

raviudari01:


src/SFA.DAS.Roatp.CourseManagement/Application/Standards/Queries/GetProviderCourse/GetProviderCourseQueryHandler.cs
```diff
@@ -51,7 +51,11 @@ public async Task<GetProviderCourseResult> Handle(GetProviderCourse.GetProviderC
                 Level = standard.Level,
                 Version = standard.Version,
                 RegulatorName = standard.ApprovalBody,
-                Sector = standard.Route
+                Sector = standard.Route,
```
DevenKShah, line 54: Convert this to use implicit operator and add tests

MarkFCain (approved):
approving, assuming you do the change Deven has requested

PR #722, opened 2022-05-19
Csp 117 get providers from roatp service

MarkFCain:


MarkFCain:
lifted into a clean branch, no longer required

PR #723, opened 2022-05-20
CSP-117 get providers from roatpservice code and coverage

MarkFCain:


src/SFA.DAS.Roatp.CourseManagement/InnerApi/Models/RegisteredProvider/RegisteredProviderModel.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace SFA.DAS.Roatp.CourseManagement.InnerApi.Models.RegisteredProvider
+{
+    public class RegisteredProviderModel
+    {
+        public int ukprn { get; set; }
```
DevenKShah, line 7: Case

src/SFA.DAS.Roatp.CourseManagement/InnerApi/Models/RegisteredProvider/Feedback.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace SFA.DAS.Roatp.CourseManagement.InnerApi.Models.RegisteredProvider
+{
+    public class Feedback
```
DevenKShah, line 5: Think we should use proper name EmployerFeedback

DevenKShah (approved):

src/SFA.DAS.Roatp.CourseManagement/InnerApi/Models/RegisteredProvider/Feedback.cs
MarkFCain: I have used the names that come back from roatp-service exactly, so diverging will cause us mapping and mapping back issues if we diverge, I suspect.  I'll leave it as is.

src/SFA.DAS.Roatp.CourseManagement/InnerApi/Models/RegisteredProvider/RegisteredProviderModel.cs
MarkFCain: yep

DevenKShah (approved):

MarkFCain:
merge update issues, deleting to create a fresh version

PR #736, opened 2022-06-08
QF-247 apprentice notifications preferences

SarahDuncan:


src/SFA.DAS.ApprenticeFeedback/Application/Queries/GetApprentice/GetApprenticeQueryHandler.cs
```diff
@@ -21,7 +21,11 @@ public async Task<GetApprenticeResult> Handle(GetApprenticeQuery request, Cancel
         {
             var result = await _apiClient.Get<GetApprenticeResponse>(new GetApprenticeRequest(request.ApprenticeId));
 
-            if (result == null)
+            var apprenticePreferences =
```
CodeWolf86, line 24: You can tweak this to be slightly more efficient, rather than awaiting each call, you can do a Task.WhenAll, and still end up with the same result.

src/SFA.DAS.ApprenticeFeedback/Application/Queries/GetApprentice/GetApprenticeQueryHandler.cs
```diff
@@ -33,6 +37,7 @@ public async Task<GetApprenticeResult> Handle(GetApprenticeQuery request, Cancel
                 Email = result.Email,
                 TermsOfUseAccepted = result.TermsOfUseAccepted,
                 ReacceptTermsOfUseRequired = result.ReacceptTermsOfUseRequired,
+                ApprenticePreferences = apprenticePreferences.ApprenticePreferences
```
CodeWolf86, line 40: This *could* be an empty list, it shouldn't be, but if it was the deriving page that shows them dynamically would just be empty.

Would potentially be more resilient, what do you think?

CodeWolf86 (changes requested):

src/SFA.DAS.ApprenticeFeedback.Api/appsettings.json
```diff
@@ -9,7 +9,7 @@
   "AllowedHosts": "*",
   "ConfigurationStorageConnectionString": "UseDevelopmentStorage=true;",
   "ConfigNames": "SFA.DAS.ApprenticeFeedback.OuterApi",
-  "Environment": "LOCAL",
+  "Environment": "DEV",
```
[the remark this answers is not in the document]
SarahDuncan: No, will change that back.

src/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetApprenticePreferencesRequest.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.SharedOuterApi.InnerApi.Requests
```
CodeWolf86, line 4: So I know it was convention, but steps are being made to begin breaking out this APIM repository as it has become far too big. the split into folders with individual pipelines has been the start of that but it can't be split out because some of them share this Outer Api project.

I would argue that apprentice preferences is pretty specific, So i'd be tempted to move this back into specifically the apprentice feedback outer api project to not give us more tech debt later on.

You may also have issues with this PR now that the master branch has shifted under you, it can be rebased to work but you may find it easier to re-branch and reapply your changes depending on your preference

src/SFA.DAS.SharedOuterApi/InnerApi/Responses/GetApprenticePreferencesResponse.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace SFA.DAS.SharedOuterApi.InnerApi.Responses
+{
+    public class GetApprenticePreferencesResponse
```
CodeWolf86, line 6: Same here for moving to app feedback out of shared outer

CodeWolf86 (changes requested):

CodeWolf86 (approved):

SarahDuncan:
Superseded by https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/796

PR #781, opened 2022-07-18
CSP-48 Implements address lookup endpoint

DevenKShah:


MarkFCain (approved):

src/Shared/SFA.DAS.SharedOuterApi/Services/LocationLookupService.cs
```diff
@@ -79,5 +79,12 @@ public async Task<LocationItem> GetLocationInformation(string location, double l
                 ? new LocationItem(location, getLocationsListItem.Location.GeoPoint, getLocationsListItem.Country) 
                 : null;
         }
+
+        public async Task<GetAddressesListResponse> GetExactMatchAddresses(string fullPostcode)
+        {
+            if (!Regex.IsMatch(fullPostcode, PostcodeRegex)) return null;
+
+            return await _locationApiClient.Get<GetAddressesListResponse>(new GetAddressesQueryRequest(fullPostcode, 1));
```
MarkFCain, line 87: please change the 1 to a suitable variable 'MinMatch' or similar

DevenKShah: Done

MarkFCain (approved):

src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.UnitTests/Application/AddressLookup/AddressLookupQueryHandlerTests.cs
```diff
@@ -0,0 +1,33 @@
...
+            var mockService = new Mock<ILocationLookupService>();
+            mockService.Setup(m => m.GetExactMatchAddresses(postcode)).ReturnsAsync(data);
+            var sut = new AddressLookupQueryHandler(mockService.Object, Mock.Of<ILogger<AddressLookupQueryHandler>>());
+
+            var result = await sut.Handle(new AddresssLookupQuery(postcode), new CancellationToken());
+
+            mockService.VerifyAll();
+            result.Addresses.Should().BeEquivalentTo(expectedAddresses);
+        }
+    }
```
MarkFCain, line 32: Add a test for null result please

MarkFCain (dismissed):
approved,request for a bit of extra coverage, but codebase is fine

src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.UnitTests/Application/AddressLookup/AddressLookupQueryHandlerTests.cs
DevenKShah: The result is never null as long as the API call is a success. If there are no results, the list is empty.

DevenKShah: Looking deeper, the service is using the dreaded Get method which supresses bad response. I will handle null.

DevenKShah: @MarkFCain I have updated the service to throw exception in case the response was not a success. Also the service returns null if the postcode does not match the regex, for this the handler just forwards null response to the controller which in turn returns a bad request.

MarkFCain (dismissed):

MarkFCain (approved):

MarkFCain (approved):

MarkFCain (dismissed):

PR #795, opened 2022-08-01
CSP-29 Implements get available standards for provider endpoint

DevenKShah:


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.Api/Controllers/ProviderCoursesController.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            return Ok(providerCourseResult);
+        }
+
+        [HttpGet]
+        [Route("providers/{ukprn}/courses")]
+        public async Task<IActionResult> GetAllProviderCourses([FromRoute] int ukprn)
+        {
+            if (ukprn <= 0)
```
MarkFCain, line 55: This should use same criteria as in GetProviderCourse ie  if (ukprn <= 9999999) and test coverage should reflect this

MarkFCain (approved):

PR #821, opened 2022-08-15
PA-441 handle sandbox mode

Pauljgraham:


src/TrackProgress/SFA.DAS.TrackProgress/Application/Commands/TrackProgressCommand.cs
```diff
@@ -59,6 +59,13 @@ public async Task<TrackProgressResponse> Handle(TrackProgressCommand request, Ca
         if (apprenticeshipResult.Body.Apprenticeships?.Count(x => x.StartDate == request.PlannedStartDate) > 1)
             return new TrackProgressResponse(HttpStatusCode.NotFound, "Multiple results for start date");
 
+        if (request.ProviderContext.InSandboxMode)
```
benagain, line 62: Ideally we could have cross-cutting concern that ensures that any changes made to the database are rolled back when in sandbox mode.  

I would be nervous about relying on remembering to handle sandbox mode, it's all to easy to forget, or modify an entity without realising.

Pauljgraham: I hadn't even considered approaching the problem that way. So are we considering forwarding the IsSandbox header to the inner API? And then automatically rolling back any transaction if IsSandbox is true?

src/TrackProgress/SFA.DAS.TrackProgress/SubscriptionHeaderConstants.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SFA.DAS.TrackProgress
+{
+    public static class SubscriptionHeaderConstants
+    {
+        public const string HeaderForProviderId = "x-request-context-subscription-name";
+        public const string HeaderForSandboxMode = "x-request-context-subscription-is-sandbox";
```
Chazzysb, line 6: Would these strings read better if they were just ProviderId and SandboxMode, without the 'HeaderFor' prefix? - the class name identifies that they are headers.. so
[FromHeader(Name = SubscriptionHeaderConstants.ProviderId)]
instead of
[FromHeader(Name = SubscriptionHeaderConstants.HeaderForProviderId)]

Pauljgraham: yes, I agree

Chazzysb (approved):
Approved

PR #822, opened 2022-08-16
Pa 435 track progress multiple validation

Chazzysb:
Pa 435 track progress multiple validation


src/TrackProgress/SFA.DAS.TrackProgress/Application/Commands/TrackProgressCommand.cs
```diff
@@ -56,8 +57,14 @@ public async Task<TrackProgressResponse> Handle(TrackProgressCommand request, Ca
...
-        if (apprenticeshipResult.Body.Apprenticeships?.Count(x => x.StartDate == request.PlannedStartDate) > 1)
-            return new TrackProgressResponse(HttpStatusCode.NotFound, "Multiple results for start date");
+        if (apprenticeshipResult.Body.TotalApprenticeshipsFound > 1)
+            apprenticeshipResult.Body.Apprenticeships?.RemoveAll(x => x.StartDate == x.StopDate);
+
+        if (apprenticeshipResult.Body.Apprenticeships?.Count == 0)
+            return new TrackProgressResponse(HttpStatusCode.NotFound, "Apprenticeship not found");
+
+        if (apprenticeshipResult.Body.Apprenticeships?.Count > 1)
+            return new TrackProgressResponse(HttpStatusCode.BadRequest, "Multiple apprenticeship records exist");
```
Pauljgraham, line 67: I think the idea of the error conditions was to return an Error object, which would be different from the Success Object. 

The error object will contain a list of errors (possibly, field, code and description). This is because we will very soon be returning multiple errors. however I'll let Ben make the call on this as for now we could get away with this pattern   

A nice example of returning different types from a controller action is here

https://github.com/SkillsFundingAgency/das-apim-endpoints/blob/master/src/VacanciesManage/SFA.DAS.VacanciesManage.Api/Controllers/VacancyController.cs#L37-L40

Pauljgraham (commented):
One comment about the return type

src/TrackProgress/SFA.DAS.TrackProgress/Application/Commands/TrackProgressCommand.cs
```diff
@@ -14,24 +14,47 @@ public record TrackProgressCommand(
...
     public string Message { get; set; } = string.Empty;
+    public string ErrorTitle { get; set; } = string.Empty;
+    public List<ErrorDetail> ErrorDetails { get; set; } = new List<ErrorDetail>();
 
     public IActionResult Result
         => StatusCode switch
         {
             HttpStatusCode.Created => new CreatedResult(string.Empty, null),
             HttpStatusCode.NotFound => new NotFoundObjectResult(Message),
+            HttpStatusCode.BadRequest => new BadRequestObjectResult(GetProblemDetails()),
```
Pauljgraham, line 34: Couple of tiny comments, not directly about your code. NotFound doesn't need any message content. If you add it the consuming applications (depending on how they are written may have to handle it). 

The second point is the error raised by the commitments search is being exposed to the consumer (this would be viewed as a security hole). If the commitments search fails. It would probably be better to log the failure and return a 405

Chazzysb: Agree, changed to 500 internal server error in this instance

src/TrackProgress/SFA.DAS.TrackProgress/Application/Commands/TrackProgressCommand.cs
Chazzysb: Understood, now returns ProblemDetails which satisfies the Jira

Pauljgraham (approved):

Chazzysb:
Changes are in 428

PR #823, opened 2022-08-16
MAC-89 - infrastructure for the Employer Accounts Outer API

VasanthaKasirajan3008:


src/EmployerAccounts/SFA.DAS.EmployerAccounts.Api/Controllers/PingController.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace SFA.DAS.EmployerAccounts.Api.Controllers
+{
+    [ApiController]
+    public class PingController : ControllerBase
+    {
+        [HttpGet]
+        [Route("pingouterapi")]
```
ChrisJWoodcock, line 9: should be just "ping"

src/EmployerAccounts/SFA.DAS.EmployerAccounts.Api/Startup.cs
```diff
@@ -0,0 +1,106 @@
...
+                    if (!_configuration.IsLocalOrDev())
+                    {
+                        o.Filters.Add(new AuthorizeFilter("default"));
+                    }
+                }).SetCompatibilityVersion(CompatibilityVersion.Version_3_0);
+
+            if (_configuration["Environment"] != "DEV")
+            {
+                services.AddHealthChecks()
+                    .AddCheck<AccountsApiHealthCheck>("Accounts API health check");
```
ChrisJWoodcock, line 64: Possibly confusing?

ChrisJWoodcock (changes requested):

src/EmployerAccounts/SFA.DAS.EmployerAccounts.Api/AppStart/AddConfigurationOptionsExtension.cs
```diff
@@ -0,0 +1,20 @@
...
+using SFA.DAS.SharedOuterApi.Configuration;
+
+namespace SFA.DAS.EmployerAccounts.Api.AppStart
+{
+    public static class AddConfigurationOptionsExtension
+    {
+        public static void AddConfigurationOptions(this IServiceCollection services, IConfiguration configuration)
+        {
+            services.AddOptions();
+            services.Configure<AccountsConfiguration>(configuration.GetSection(nameof(AccountsConfiguration)));
```
ChrisJWoodcock, line 14: This should be:
services.Configure<AccountsConfiguration>(configuration.GetSection("AccountsApiConfiguration"));

as the configuration file is called "AccountsApiConfiguration" where as the class name is AccountsConfiguration

ChrisJWoodcock (changes requested):

src/EmployerAccounts/SFA.DAS.EmployerAccounts.Api/AppStart/AddConfigurationOptionsExtension.cs
ChrisJWoodcock: Actually I was wrong yours was consistent between the code and config originally! - I will make it more consistent with other outerapi's in my branch

ChrisJWoodcock (approved):

PR #825, opened 2022-08-17
CSP-145-Outer-DeleteStandard

raviudari01:


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.Api.UnitTests/Controllers/ProviderCourseDeleteControllerTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        [Test, MoqAutoData]
+        public async Task DeleteProviderCourse_InvokesCommand(
+            [Frozen] Mock<IMediator> mediatorMock,
+            [Greedy] ProviderCourseDeleteController sut,
+            int ukprn, int larsCode, DeleteProviderCourseCommand command)
+        {
+            await sut.DeleteProviderCourse(ukprn, larsCode, command);
+
+            mediatorMock.Verify(m => m.Send(command, It.IsAny<CancellationToken>()));
```
MarkFCain, line 26: You could combine this line into the test below. I don't mind it staying as is though, I'll leave the decision to you.

MarkFCain (dismissed):

src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.Api.UnitTests/Controllers/ProviderCourseDeleteControllerTests.cs
DevenKShah: One assert per test is best practise that we have not been following.

DevenKShah (approved):

DevenKShah (approved):

PR #844, opened 2022-08-31
Pa 470 show api on dev portal

Chazzysb:
Pa 470 show api on dev portal


Chazzysb:
Closed as this work is done under 469

PR #845, opened 2022-09-01
PA-469 add api swagger content 

Pauljgraham:


src/TrackProgress/SFA.DAS.TrackProgress.Api/Controllers/TrackProgressController.cs
```diff
@@ -36,6 +36,24 @@ public string Ukprn
...
+    ///
+    /// The overall progress of this apprenticeship will be constructed from these submissions. 
+    /// </remarks>
+    /// <param name="uln">The apprentice's Unique Learner Number.</param>
+    /// <param name="plannedStartDate">The planned start date for this apprenticeship.</param>
+    /// <param name="progress">The taxonomy content.Accepts an array of KSB progress percentiles.
+    /// 
+    /// The "id" field must be the GUID identifier of the KSB for the apprenticeship's course.
+    /// 
+    /// The "value" field must be the percentile range 1-100, denoting what percentage of the KSB has been completed.
```
benagain, line 54: In the last few hours the range has been changed to 0-100 (not 1-100).

However, there will be other places that need to be updated, so let's do them all together in a different PR.

benagain (dismissed):

src/TrackProgress/SFA.DAS.TrackProgress.Api/Controllers/TrackProgressController.cs
Pauljgraham: OK, I'll create one if you like

Chazzysb (approved):
Approved

Chazzysb (approved):
Approved

PR #893, opened 2022-10-07
QF-493-Calculation

SarahDuncan:


src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetStandardRequest.cs
```diff
@@ -4,12 +4,12 @@ namespace SFA.DAS.SharedOuterApi.InnerApi.Requests
 {
     public class GetStandardRequest : IGetApiRequest
     {
-        public GetStandardRequest(int id)
+        public GetStandardRequest(string standardUId)
```
CodeWolf86, line 7: I'd be tempted to move this request into the campaigns project.

My concern is this is in the Shared Outer Api project, and other Outer Api's may be using this with an "int" parameter, but we don't see that in the solution split.

Your change is the right one, i'm just thinking about making less waves. So i'd add this to the campaigns requests area with a string and same url, leaving the shared outer api alone, the same with the GetAvailableToStartStandardsListRequest file which had an additional using.

If we can take out the two shared changes, it means it won't kick off every shared outer api build on completion

Also the request string can be a LarsCode, IFate Reference Number or StandardUId, as the endpoint can understand all three. So a more specific request in our project might also make more sense.

CodeWolf86 (changes requested):
Was this meant to be deleted? src/Campaign/SFA.DAS.Campaign.UnitTests/SFA.DAS.Campaign.UnitTests.csproj

CodeWolf86 (approved):

PR #903, opened 2022-10-18
CSP-278 Refactored post shortlist to talk to shortlist api

DevenKShah:


src/FindApprenticeshipTraining/SFA.DAS.FindApprenticeshipTraining/Services/ShortlistApiClient.cs
```diff
@@ -0,0 +1,86 @@
...
+        
+        public Task Put(IPutApiRequest request)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task Put<TData>(IPutApiRequest<TData> request)
+        {
+            throw new NotImplementedException();
+        }
```
MarkFCain, line 84: Are all these needed? If they are, because it's inheriting from IInternalApiClient, ignore this comment and mark as resolved

MarkFCain (approved):

src/FindApprenticeshipTraining/SFA.DAS.FindApprenticeshipTraining/Services/ShortlistApiClient.cs
DevenKShah: Ignoring your comment :-)

PR #916, opened 2022-10-28
CSP-287 provider details endpoint

DevenKShah:


src/FindApprenticeshipTraining/SFA.DAS.FindApprenticeshipTraining/InnerApi/Responses/CouseLocationModel.cs
```diff
@@ -13,6 +14,7 @@ public class CouseLocationModel
         public double? Longitude { get; set; }
     }
 
+    //MFCMFC this is probably going
```
DevenKShah: I presume these are notes for the next work, make sure none of these are left behind. Also suggest make notes out of the code.

src/FindApprenticeshipTraining/SFA.DAS.FindApprenticeshipTraining/Application/TrainingCourses/Queries/GetTrainingCourseProvider/GetTrainingCourseProviderQueryHandler.cs
```diff
@@ -1,134 +1,162 @@
...
-                    providerTask.Result.DeliveryTypes = new List<GetDeliveryTypeItem> 
-                    {
-                        new GetDeliveryTypeItem
-                        {
-                            DeliveryModes = "NotFound"
-                        } 
-                    };    
-                }
-            }
+            var providerDetails = await GetProviderDetails(request.ProviderId, request.CourseId, locationTask.Result, shortlistTask.Result);
```
DevenKShah, line 67: This can be included in the parallel task as well.

MarkFCain: No it cannot, as it requires the result from shortlistTask

PR #931, opened 2022-11-17
CSP-384 Added Roatp Provider Moderation outer api solution and projects

DevenKShah:


src/RoatpProviderModeration/SFA.DAS.RoatpProviderModeration.Application/Queries/GetProvider/GetProviderQuery.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace SFA.DAS.RoatpProviderModeration.Application.Queries.GetProvider
+{
+    public class GetProviderQuery : IRequest<GetProviderQueryResponse>
+    {
+        public int Ukprn { get; set; }
+    }
+
+    public class GetProviderQueryResponse
+    {
+
```
MarkFCain, line 12: empty space

DevenKShah: Will sort it when it gets implemented... just a place holder for now

PR #933, opened 2022-11-17
CSP-384 Create Roatp Provider Moderation solution and projects

DevenKShah:


jack-cloud-platform (approved):
Approving for pipeline file addition

PR #950, opened 2022-11-29
PA-592-UpgradeToDotNetSix

Chazzysb:
PA-592-UpgradeToDotNetSix


alireid (changes requested):
Hiya Chas, I get a build error relating to downgrade of MediatR in included projects, UnitTests being one of them, in addition can we look to upgrade some of the internal references?

alireid:
Given this apim sub solution calls in other artefacts, namely SFA.DAS.SharedOuterApi, having spoken with Chas and seen other teams upgrades it will make it very awkward to do a blanket update on other nuget packages at this stage. Therefore this can be merged in its current upgraded state.

alireid (approved):

PR #956, opened 2022-11-30
CSP-309 updated configuration to use roatpv2 instead of courseDelivery

MarkFCain:


src/EmployerDemand/SFA.DAS.EmployerDemand.Api/AppStart/AddConfigurationOptionsExtension.cs
```diff
@@ -13,8 +13,8 @@ public static void AddConfigurationOptions(this IServiceCollection services, ICo
             services.AddOptions();
             services.Configure<CoursesApiConfiguration>(configuration.GetSection(nameof(CoursesApiConfiguration)));
             services.AddSingleton(cfg => cfg.GetService<IOptions<CoursesApiConfiguration>>().Value);
-            services.Configure<CourseDeliveryApiConfiguration>(configuration.GetSection(nameof(CourseDeliveryApiConfiguration)));
-            services.AddSingleton(cfg => cfg.GetService<IOptions<CourseDeliveryApiConfiguration>>().Value);
+            services.Configure<RoatpV2ApiConfiguration>(configuration.GetSection(nameof(RoatpV2ApiConfiguration)));
+            services.AddSingleton(cfg => cfg.GetService<IOptions<RoatpV2ApiConfiguration>>().Value);
```
DevenKShah, line 17: If there is no more dependency on CourseDelivery, then we should remove the config for this. But make sure that config does not get merged, we will have to keep that in a FATIntegration Branch in emlpoyer config repo.

DevenKShah (approved):
Approved with a config change recommendation

src/EmployerDemand/SFA.DAS.EmployerDemand.Api/AppStart/AddConfigurationOptionsExtension.cs
MarkFCain: Ok I'll do that

PR #988, opened 2023-01-25
CSP-544 all providers endpoint changed from course delivery to Roatp V2

MarkFCain:
…roatp course management

### DIFF
--- a/src/Recruit/SFA.DAS.Recruit.Api/Models/GetProviderAddress.cs
+++ b/src/Recruit/SFA.DAS.Recruit.Api/Models/GetProviderAddress.cs
@@ -13,14 +13,17 @@ namespace SFA.DAS.Recruit.Api.Models
 
         public static implicit operator GetProviderAddress(GetProvidersListItemAddress source)
         {
+            if (source==null)
+                return new GetProviderAddress();
+            
             return new GetProviderAddress
             {
-                Address1 = source?.AddressLine1,
-                Address2 = source?.AddressLine2,
-                Address3 = source?.AddressLine3,
-                Address4 = source?.AddressLine4,
-                Town = source?.Town,
-                Postcode = source?.Postcode
+                Address1 = source.AddressLine1,
+                Address2 = source.AddressLine2,
+                Address3 = source.AddressLine3,
+                Address4 = source.AddressLine4,
+                Town = source.Town,
+                Postcode = source.Postcode
             };
         }
     }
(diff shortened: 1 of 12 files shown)


src/Recruit/SFA.DAS.Recruit.Api/Models/GetProviderAddress.cs
```diff
@@ -15,12 +15,12 @@ public class GetProviderAddress
         {
             return new GetProviderAddress
             {
-                Address1 = source.Address1,
-                Address2 = source.Address2,
-                Address3 = source.Address3,
-                Address4 = source.Address4,
-                Town = source.Town,
-                Postcode = source.Postcode
+                Address1 = source?.AddressLine1,
```
DevenKShah, line 18: Can source be null ever? if it can be can we check that first instead of having null check on each assignment?

src/Recruit/SFA.DAS.Recruit.sln
```diff
@@ -42,9 +42,8 @@ Global
 		{D5D4AB58-5E91-42DF-ABFF-E03BCDCD83DB}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{D5D4AB58-5E91-42DF-ABFF-E03BCDCD83DB}.Release|Any CPU.Build.0 = Release|Any CPU
 		{41D340D7-A7E7-4430-B7F3-C5C1A90DD113}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-        {41D340D7-A7E7-4430-B7F3-C5C1A90DD113}.Debug|Any CPU.Build.0 = Debug|Any CPU
-        {41D340D7-A7E7-4430-B7F3-C5C1A90DD113}.Release|Any CPU.ActiveCfg = Release|Any CPU
```
DevenKShah, line 46: Is this change required?

DevenKShah (approved):
Approved with minor code improvement

src/Recruit/SFA.DAS.Recruit.Api/Models/GetProviderAddress.cs
MarkFCain: Yes, there are one or two providers out of the hundreds with no address details.  I'll have a look and rework if needed.

src/Recruit/SFA.DAS.Recruit.sln
MarkFCain: no, I'll revert.  VS did that, not me...

PR #993, opened 2023-02-03
198 get apprentice names api

dinathain:
Added new client, service etc. for for fetching all apprenticeships from the Apprenticeship inner API (from the Funding Outer API)


dinathain:
Closed due to renaming branch

PR #999, opened 2023-02-07
CSP-500 - AanHub Outer Api GET /Regions

LiamWright:
Initial endpoint for ApprenticeAan endpoint.

Regions controller/handler implemented as well as base classes needed for successful API calls.

ApprenticeAan also had not been added to the parent project, which has since been done now.


src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Program.cs
```diff
@@ -42,10 +44,6 @@
 
 // Configure the HTTP request pipeline.
 
-if (app.Environment.IsDevelopment())
```
DevenKShah, line 45: Was this removed on purpose?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQuery.cs
```diff
@@ -0,0 +1,8 @@
+﻿using MediatR;
+
+namespace SFA.DAS.ApprenticeAan.Application.Regions.Queries.GetRegions
```
DevenKShah, line 3: This should be along side the Query Handler.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            try
+            {
+                var result = await _mediator.Send(new GetRegionsQuery());
+
+                if (result == null) return NotFound();
+
+                return Ok(result);
+            }
+            catch { return BadRequest(); }
```
DevenKShah, line 33: Remove try catch. Any errors should just bubble up and result in 500 response.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public RegionsController(IMediator mediator) => _mediator = mediator;
+
+        /// <summary>
+        ///     Get list of regions
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status204NoContent)]
```
DevenKShah, line 21: Remove 204 response type.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
```diff
@@ -0,0 +1,36 @@
...
+        [ProducesResponseType(typeof(GetRegionsQueryResult), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetRegions()
+        {
+            try
+            {
+                var result = await _mediator.Send(new GetRegionsQuery());
+
+                if (result == null) return NotFound();
```
DevenKShah, line 29: I would return a 500 here as the expectations is to always get some data back. Also it is possible that inner API returned a 500.

DevenKShah (changes requested):
We are trying to minimize the dependency on the shared projects hence I had the ApiClient bits in Api project. All these needs to be moved back please. There should be no changes to the shared projects.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Program.cs
LiamWright: Yes, SONAR flagged this as a security hotspot

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQuery.cs
LiamWright: I'm not sure what you mean by this. `GetRegionsQuery` is in the same folder. Does it need to be in the same class?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
LiamWright: Done

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
LiamWright: Done

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/RegionsController.cs
LiamWright: Implemented

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Program.cs
DevenKShah: That cannot be right, we do need to see exception page when debugging.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Common/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace SFA.DAS.ApprenticeAan.Application.Common
+{
+    public static class Constants
+    {
+        public static class AanHubApiUrls
+        {
+            private const string ApiPrefix = "/api";
+            public const string GetRegionsUrl = $"{ApiPrefix}/regions";
```
DevenKShah, line 8: Don't see a point of having specific urls as constants. This pattern will become complicated when you have dynamic routes to build.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQuery.cs
DevenKShah: duh moment.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQueryResult.cs
```diff
@@ -0,0 +1,14 @@
+﻿using SFA.DAS.ApprenticeAan.Application.Entities;
+
+namespace SFA.DAS.ApprenticeAan.Application.Regions.Queries.GetRegions
+{
+    public class GetRegionsQueryResult
```
DevenKShah, line 5: We dont have to have Result object if it does not make sense, like in this case the Response can easily be List<Region>

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application.UnitTests/Regions/Queries/GetRegions/GetRegionsQueryHandlerTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            GetRegionsQuery query,
+            [Frozen] Mock<IAanHubApiClient<AanHubApiConfiguration>> apiClient,
+            [Frozen(Matching.ImplementedInterfaces)]
+            GetRegionsQueryHandler handler)
+        {
+            apiClient.Setup(x =>
+                    x.GetWithResponseCode<GetRegionsQueryResult>(It.IsAny<GetRegionsQueryRequest>()))
+                .ReturnsAsync(new ApiResponse<GetRegionsQueryResult>(null!, HttpStatusCode.BadRequest, string.Empty));
+
+            var result = await handler.Handle(query, CancellationToken.None);
```
DevenKShah, line 47: We need a way to communicate non success response further up so web consumer can react accordingly. This one I am not too bothered as we have static data.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Program.cs
LiamWright: SONAR cloud doesn't keep the logs, so this historical run is the most detailed report I can get you, but it did in fact flag it.

image


Not sure what's changed with it now, but I've reinstated it.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQueryHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private readonly IAanHubApiClient<AanHubApiConfiguration> _apiClient;
+
+        public GetRegionsQueryHandler(IAanHubApiClient<AanHubApiConfiguration> apiClient) => _apiClient = apiClient;
+
+        public async Task<GetRegionsQueryResult?> Handle(GetRegionsQuery request, CancellationToken cancellationToken)
+        {
+            var regionsResult = await _apiClient.GetWithResponseCode<GetRegionsQueryResult>(new GetRegionsQueryRequest());
+
+            if (regionsResult.StatusCode != HttpStatusCode.OK || regionsResult.Body == null) return null;
```
DevenKShah, line 19: Need to add a log error for unsuccessful response here as we do not expect this to not return data.

DevenKShah (approved):
Few more minor suggestions. I am approving as it all looks good otherwise.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQueryResult.cs
LiamWright: No, I thought that. But we might as well keep it mirrored to AanHub.Api if it's not too much overhead?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQueryHandler.cs
LiamWright: Added logError for unsuccessful calls

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Common/Constants.cs
DevenKShah: IMHO it is just unnecessary hop in the code to check the URL. I am expecting that one request will have one endpoint to associate with, so dont think there is any benefit. If we find a benefit of doing this down the line then we can always refactor.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Regions/Queries/GetRegions/GetRegionsQueryResult.cs
DevenKShah: I am not too fussed, but I would rather avoid any boiler plate code.

DevenKShah (dismissed):

MarkFCain (approved):

PR #1010, opened 2023-02-24
CSP-593 GET provider course locations endpoint changed to not raise e…

MarkFCain:
…rror if empty set returned


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.Api/Controllers/ProviderCourseLocationsController.cs
```diff
@@ -25,7 +27,7 @@ public async Task<IActionResult> GetProviderCourseLocations(int ukprn, int larsC
         {
             _logger.LogInformation("Outer API: Request received to get provider course locations for ukprn: {ukprn} larscode: {larscode}", ukprn, larsCode);
             var providerCourselocationsResult = await _mediator.Send(new GetProviderCourseLocationQuery(ukprn, larsCode));
-            if (providerCourselocationsResult == null || !providerCourselocationsResult.ProviderCourseLocations.Any())
+            if (providerCourselocationsResult == null)
             {
                 _logger.LogError($"Provider Course Locations not found for ukprn {ukprn} and lars code {larsCode}");
```
DevenKShah, line 32: Drop the log level to information.

PR #1015, opened 2023-03-02
FLP-233 Don't show editable summary in change of provider scenario

tom-gough:


dinathain:
@ShakilUmar @tom-gough  

Closing this PR since we are choosing a slightly simpler implementation which will instead use the IsContinuation field (this should encompass the change of provider/employer journeys and not require this new field to be passed around). 

See updated implementation to provider commitments here: https://github.com/SkillsFundingAgency/das-providercommitments/pull/493
and employer commitments here: https://github.com/SkillsFundingAgency/das-employercommitments-v2/pull/410

PR #1020, opened 2023-03-08
Flp 203 add draft apprenticeship endpoints

MichaelYoung1981:


src/Approvals/SFA.DAS.Approvals.UnitTests/Application/Cohorts/Commands/CreateCohortCommandHandlerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        public void Setup()
+        {
+            _fixture = new Fixture();
+            _request = _fixture.Create<CreateCohortCommand>();
+
+            _commitmentsApiClient = new Mock<ICommitmentsV2ApiClient<CommitmentsV2ApiConfiguration>>();
+
+            _handler = new CreateCohortCommandHandler(_commitmentsApiClient.Object);
+        }
+
```
dinathain, line 35: Do we want to add some non-happy path tests in here in the case that the response from commitments is null (ArgumentNullException thrown) or unsuccessful (ApiResponseException thrown)?

src/Approvals/SFA.DAS.Approvals.UnitTests/Application/DraftApprenticeships/Commands/AddDraftApprenticeshipCommandHandlerTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        public void Setup()
+        {
+            _fixture = new Fixture();
+            _request = _fixture.Create<AddDraftApprenticeshipCommand>();
+
+            _commitmentsApiClient = new Mock<ICommitmentsV2ApiClient<CommitmentsV2ApiConfiguration>>();
+
+            _handler = new AddDraftApprenticeshipCommandHandler(_commitmentsApiClient.Object);
+        }
+
```
dinathain, line 35: Same here (for null/unsuccessful responses)

src/Approvals/SFA.DAS.Approvals.UnitTests/Application/DraftApprenticeships/Commands/UpdateDraftApprenticeshipCommandHandlerTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        public void Setup()
+        {
+            _fixture = new Fixture();
+            _request = _fixture.Create<UpdateDraftApprenticeshipCommand>();
+
+            _commitmentsApiClient = new Mock<ICommitmentsV2ApiClient<CommitmentsV2ApiConfiguration>>();
+
+            _handler = new UpdateDraftApprenticeshipCommandHandler(_commitmentsApiClient.Object);
+        }
+
```
dinathain, line 36: Same again here (for null/unsuccessful responses)

dinathain (changes requested):

src/Approvals/SFA.DAS.Approvals/InnerApi/Requests/AddDraftApprenticeshipRequest.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace SFA.DAS.Approvals.InnerApi.Requests
+{
+    public class AddDraftApprenticeshipRequest
+    {
+        public UserInfo UserInfo { get; set; }
```
dinathain, line 7: Also noted that all of these request objects are very similar (I _think_ the same?) - can we use abstraction to reduce code duplication?

MichaelYoung1981: I've had this argument many times with many people and the counter argument, especially when it's APIM, is around allowing the objects to diverge independently.  Personally, given the amount of copy/paste I've had to do here and previous experience, I think there is definitely something wrong with the overall design of our outer APIs but I've thus far not been able to build a significant enough argument against.

dinathain: Fair enough - I guess this ensures consistency with the rest of the code-base until a decision is (potentially) made otherwise. Thanks for explaining the position.

dinathain (approved):

PR #1022, opened 2023-03-10
CSP-499 Implements Get Addresses endpoint

DevenKShah:


src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace SFA.DAS.ApprenticeAan.Api.UnitTests.Controllers;
+
+public class LocationsControllerTests
+{
+    [Test]
+    [MoqAutoData]
+    public async Task GetAddresses_ReturnsAddresses(
+        GetAddressesQueryResult response,
+        [Frozen] Mock<IMediator> mockMediator,
```
ravihejmadi, line 18: The mockMediator and sut parameters could go into a ctor as seen [here](https://github.com/SkillsFundingAgency/das-aan-hub-api/blob/b5edde3a8e2081e9738a6cad615e9869da408d03/src/SFA.DAS.AANHub.Api.UnitTests/Controllers/AdminsControllerTests.cs#L30)

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        result.As<OkObjectResult>().Value.Should().BeEquivalentTo(response);
+    }
+
+    [Test]
+    [MoqAutoData]
+    public async Task GetAddresses_NoMatch_ReturnNotFoundResponse(
+        [Frozen] Mock<IMediator> mockMediator,
+        [Greedy] LocationsController sut)
+    {
+        var response = new GetAddressesQueryResult();
```
YuvrajPatil-DFE, line 38: Minor: Creation of response object can be simplified using MoqAutoData used in previous test case **Task GetAddresses_ReturnsAddresses**

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        mockMediator
+            .Setup(m => m.Send(It.IsAny<GetAddressesQuery>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(response);
+
+        var result = await sut.GetAddresses("thisIsAQuery");
+
+        result.As<OkObjectResult>().Should().NotBeNull();
+
+        result.As<OkObjectResult>().Value.Should().BeEquivalentTo(response);
```
ravihejmadi, line 29: `mediatorMock.Verify(m => m.Send(It.IsAny<GetAdminMemberQuery>(), It.IsAny<CancellationToken>()));` can go here as seen [here](https://github.com/SkillsFundingAgency/das-aan-hub-api/blob/b5edde3a8e2081e9738a6cad615e9869da408d03/src/SFA.DAS.AANHub.Api.UnitTests/Controllers/AdminsControllerTests.cs#L101)

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/LocationsController.cs
```diff
@@ -0,0 +1,27 @@
...
+    public LocationsController(IMediator mediator)
+    {
+        _mediator = mediator;
+    }
+
+    [HttpGet]
+    public async Task<IActionResult> GetAddresses([FromQuery] string query)
+    {
+        var queryResponse = await _mediator.Send(new GetAddressesQuery(query));
+
```
MarkFCain, line 22: I would suggest if (queryResponse==null) return BadRequest() here, and add coverage if you do.

MarkFCain (approved):

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Controllers/LocationsController.cs
DevenKShah: It is not expected that the query response will be null, it can be empty list if there is no match.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
DevenKShah: @ravihejmadi The admin controller test is wrong, it cannot be in ctor, it has to be in a SetUp method. Unlike in XUnit, NUnit creates a single instance of a class for the tests, which means any class level properties initialized in constructor will share the setup in all the tests.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
DevenKShah: I will add a separate test for this.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/LocationsControllerTests.cs
DevenKShah: I didn't want the property in response to be auto populated, as I am testing empty response.

MarkFCain (dismissed):

YuvrajPatil-DFE (approved):

MarkFCain (approved):

PR #1023, opened 2023-03-13
Csp 597 delete course directory and course delivery code and configuration

MarkFCain:


src/Reservations/SFA.DAS.Reservations/Application/Providers/Queries/GetProvider/GetProviderQueryHandler.cs
```diff
@@ -10,42 +10,25 @@ namespace SFA.DAS.Reservations.Application.Providers.Queries.GetProvider
...
-                        Ukprn = request.Ukprn
-                    });
-            }
+            var result = await _roatpApiClient.Get<GetRoatpProviderResponse>(
+                new GetProviderRequest
+                {
+                    Ukprn = request.Ukprn
+                });
 
+            provider = result;
```
cofaulco, line 30: this line could be removed

cofaulco (changes requested):

cofaulco (approved):

src/EmployerFinance/SFA.DAS.EmployerFinance/InnerApi/Responses/GetProviderResponse.cs
```diff
@@ -1,10 +1,10 @@
-﻿namespace SFA.DAS.SharedOuterApi.InnerApi.Responses.ProviderCourses
+﻿namespace SFA.DAS.EmployerFinance.InnerApi.Responses
 {
     public class GetProviderResponse
     {
         public int Ukprn { get; set; }
-        public string LegalName { get; set; }
-        public string Website { get; set; }
+        public string Name { get; set; }
```
DevenKShah, line 6: Is this change required?

DevenKShah (approved):

src/EmployerFinance/SFA.DAS.EmployerFinance/InnerApi/Responses/GetProviderResponse.cs
MarkFCain: it does appear to need that change.  For some reason, it was awaiting the fields 'LegalName' and 'Website' from roatp inner, but the fields are actually 'name' and 'ContactUrl', hence the change.    And Corey said this endpoint is  called from the EmployerFinance messagehandlers during the MA payment pull.   I presume such calls would get a null value for the unmapped fields at the moment.

PR #1029, opened 2023-03-22
QT1-127 Use Portal Outer API : Expose UpdateApprentice endpoint

alireid:
And endpoints for get and patch


src/ApprenticePortal/SFA.DAS.ApprenticePortal.Api/Startup.cs
```diff
@@ -36,6 +38,9 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton(_env);
 
+            services.Configure<ApprenticePortalConfiguration>(_configuration.GetSection(nameof(ApprenticePortalConfiguration)));
+            services.AddSingleton(cfg => cfg.GetService<IOptions<ApprenticePortalConfiguration>>().Value);
+
```
Pauljgraham, line 43: Don't think will be needed anymore

src/ApprenticePortal/SFA.DAS.ApprenticePortal/Application/Commands/ApprenticePatchCommand.cs
```diff
@@ -0,0 +1,13 @@
+﻿using MediatR;
+using Microsoft.AspNetCore.JsonPatch;
+using SFA.DAS.ApprenticePortal.Models;
+using System;
+
+namespace SFA.DAS.ApprenticePortal.Application.Commands.ApprenticePatch
+{
+    public class ApprenticePatchCommand : IRequest<Unit>
```
Pauljgraham, line 8: IRequest<Unit> isn't required, IRequest should be fine

Pauljgraham (commented):
Looks great, couple of tiny, tiny comments.

src/ApprenticePortal/SFA.DAS.ApprenticePortal/Application/Commands/ApprenticePatchCommand.cs
Pauljgraham: Sorry 

IRequest<Unit>

is not usually defined explicitly

IRequest (without the Unit) is fine

Pauljgraham: but I'm fine with it either way

Pauljgraham (approved):

PR #1030, opened 2023-03-23
QT1-128 populate current apprenticeship details from Commitments and Provider

Pauljgraham:


alireid (dismissed):

alireid (approved):

src/ApprenticePortal/SFA.DAS.ApprenticePortal/InnerApi/ApprenticeAccounts/Requests/PostApprenticeshipRequest.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using SFA.DAS.ApprenticePortal.Models;
+using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.ApprenticePortal.InnerApi.ApprenticeAccounts.Requests
+{
+    public class PostApprenticeshipRequest : IPostApiRequest<MyApprenticeshipData>
+    {
+        private readonly Guid _apprenticeId;
+        public string PostUrl => $"/apprentices/{_apprenticeId}/my-apprenticeship";
```
alandburns, line 10: Should be "/myapprenticeship" and not "my-apprenticeship" on the call to Apprentice Accounts Inner API.
Also as there is the possibility that there is a follow-on re-confirm of the same Apprentcieship will need to perform check GET /apprentices/{_apprenticeId}/myapprenticeship/{commitments apprentcieshipid}
and then if response includes the MyApprenticeship record then issue PUT to replace rather than POST to create.

src/ApprenticePortal/SFA.DAS.ApprenticePortal/InnerApi/ApprenticeAccounts/Requests/GetMyApprenticeshipRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class GetMyApprenticeshipRequest : IGetApiRequest
+    {
+        private readonly Guid _id;
+
+        public GetMyApprenticeshipRequest(Guid id)
+        {
+            _id = id;
+        }
+
+        public string GetUrl => $"apprentices/{_id}/my-apprenticeship";
```
alandburns, line 15: Should be "/myapprenticeship" and not "my-apprenticeship" on the call to Apprentice Accounts Inner API.

PR #1037, opened 2023-04-03
QF-856 Remove Headers from Apprentice Feedback

SarahDuncan:
Remove the 'X-Powered-By' and 'Server' headers from the Apprentice Feedback Outer API to resolve pen test issues.

### DIFF
--- a/src/ApprenticeFeedback/SFA.DAS.ApprenticeFeedback.Api/web.config
+++ b/src/ApprenticeFeedback/SFA.DAS.ApprenticeFeedback.Api/web.config
@@ -1,0 +1,13 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+	<system.webServer>
+		<security>
+			<requestFiltering removeServerHeader="true" />
+		</security>
+		<httpProtocol>
+			<customHeaders>
+				<remove name="X-Powered-By" />
+			</customHeaders>
+		</httpProtocol>
+	</system.webServer>
+</configuration>


ChrisJWoodcock (changes requested):
I think this is the wrong repo - check the ticket and url which I think is to the apprentice-feedback-api not the outer api of the similar/same name

SarahDuncan:
> I think this is the wrong repo - check the ticket and url which I think is to the apprentice-feedback-api not the outer api of the similar/same name

I believe this is the right API, because the URL (**https://pp-apprentice-feedback-api.apprenticeships.education.gov.uk/api**/apprenticefeedbacktarget/7D3BFF43-8755-4AD4-85D3-50986BC04CF6/exitsurvey) is the same as the URL in ticket [QF-1265](https://skillsfundingagency.atlassian.net/browse/QF-1265?search_id=aa7d7d9f-e485-41e0-8002-a2f996468b2c) (which was an outer API change and has been [approved](https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/1038)) with only the specific endpoint varying, and the response shown in both that ticket and this ticket include version numbers suggesting they're the same API called (so if QF-1265 is outer API I think QF-856 is too). I also think since only the outer API is exposed it seems more likely?

SarahDuncan:
> I think this is the wrong repo - check the ticket and url which I think is to the apprentice-feedback-api not the outer api of the similar/same name

> I think this is the wrong repo - check the ticket and url which I think is to the apprentice-feedback-api not the outer api of the similar/same name

I believe this is the right API, because the URL is the same as the URL in ticket [QF-1265](https://skillsfundingagency.atlassian.net/browse/QF-1265?search_id=aa7d7d9f-e485-41e0-8002-a2f996468b2c) (which was an outer API change and has been [approved](https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/1038)) with only the specific endpoint varying, and the response shown in both that ticket and this ticket include version numbers suggesting they're the same API called. I also think since only the outer API is exposed it seems more likely that pen test changes would be in the outer API?

- URL from QF-1265 **https://pp-apprentice-feedback-api.apprenticeships.education.gov.uk/api/**apprenticefeedbacktarget/7D3BFF43-8755-4AD4-85D3-50986BC04CF6/exitsurvey
- URL from QF-865 **https://pp-apprentice-feedback-api.apprenticeships.education.gov.uk/api/**attributes/feedback

SarahDuncan:
> I think this is the wrong repo - check the ticket and url which I think is to the apprentice-feedback-api not the outer api of the similar/same name

I believe this is the right API, because the URL is the same as the URL in ticket [QF-1265](https://skillsfundingagency.atlassian.net/browse/QF-1265?search_id=aa7d7d9f-e485-41e0-8002-a2f996468b2c) (which was an outer API change and has been [approved](https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/1038)) with only the specific endpoint varying, and the response shown in both that ticket and this ticket include version numbers suggesting they're the same API. I also think since only the outer API is exposed it seems more likely that pen test changes would be in the outer API?

- URL from QF-1265 https://pp-apprentice-feedback-api.apprenticeships.education.gov.uk/api/apprenticefeedbacktarget/7D3BFF43-8755-4AD4-85D3-50986BC04CF6/exitsurvey
- URL from QF-865 https://pp-apprentice-feedback-api.apprenticeships.education.gov.uk/api/attributes/feedback

[QF-1265]: https://skillsfundingagency.atlassian.net/browse/QF-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ChrisJWoodcock:
Closing this without merge as we won't do the outer API as there is no Pen testing issue raised on it

PR #1046, opened 2023-04-20
QT1-187 endpoint for priorlearningdata

alireid:


Pauljgraham (commented):
Looks good and I know what you mean by scaffolding, it's a little over the top. But we do seem to be missing some unit tests for this code

Pauljgraham (changes requested):
We appear to be missing a test on the new controller endpoint that is also mapping to the handler and should be tested

Pauljgraham (approved):

PR #1070, opened 2023-05-24
CSP-528 Employer Outer API Get Regions

YuvrajPatil-DFE:


src/EmployerAan/SFA.DAS.EmployerAan.Api/Controllers/RegionsController.cs
```diff
@@ -0,0 +1,27 @@
+﻿using MediatR;
+using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.EmployerAan.Application.Regions.Queries.GetRegions;
+
+namespace SFA.DAS.EmployerAan.Api.Controllers
```
MarkFCain, line 5: fix the namespace scope

MarkFCain (dismissed):
approved, but please fix the namespace scope

PR #1084, opened 2023-06-14
QT1-293 Cohort All Details Collation 

alireid:


alireid:
Happy with all these extra changes

Pauljgraham (dismissed):

src/Approvals/SFA.DAS.Approvals.Api/Controllers/CohortController.cs
```diff
@@ -64,11 +64,11 @@ public async Task<IActionResult> GetCohortDetails(long cohortId)
                     return NotFound();
                 }
 
-                return Ok((GetCohortDetailsResponse)result);
+                return Ok(result);
```
chrisfoster76, line 67: We've lost the conversion from result -> response here. Query results are an internal concern and should not be directly exposed as an api response.

src/Approvals/SFA.DAS.Approvals.Api/Controllers/CohortController.cs
```diff
@@ -64,11 +64,11 @@ public async Task<IActionResult> GetCohortDetails(long cohortId)
...
             }
             catch (Exception e)
             {
-                _logger.LogError(e, $"Error in Get Cohort Details - cohort id {cohortId}");
+                _logger.LogError(e, $"Error in Get All Cohort Details - cohort id {cohortId}");
```
chrisfoster76, line 71: Minor and very pedantic, but I don't quite follow the change here or what it is supposed to _mean_. The endpoint should produce all the details it needs to for the UI endpoint in context, no more, no less, so the word "All" seems redundant.

src/Approvals/SFA.DAS.Approvals/Application/Cohorts/Queries/GetCohortDetails/GetCohortDetailsQuery.cs
```diff
@@ -22,12 +22,35 @@ public class GetCohortDetailsQuery : IRequest<GetCohortDetailsQueryResult>
...
+        public ApprenticeshipEmployerType LevyStatus { get; set; }
+        public long? ChangeOfPartyRequestId { get; set; }
+        public bool IsLinkedToChangeOfPartyRequest { get; set; }
+        public TransferApprovalStatus? TransferApprovalStatus { get; set; }
+        public LastAction LastAction { get; set; }
+        public bool ApprenticeEmailIsRequired { get; set; }
         public bool HasUnavailableFlexiJobAgencyDeliveryModel { get; set; }
         public IEnumerable<string> InvalidProviderCourseCodes { get; set; }
-
+        public IEnumerable<DraftApprenticeship> DraftApprenticeships { get; set; }
```
chrisfoster76, line 51: I don't love this - it means that the query result is partly defined by the definition of an inner api type. The query result's definition should be entirely self-contained, even if that seems like a lot of mapping overhead.

But you can see the problem: if the inner api's definition changes, the query result's definition will also change, and that could have unforseen consequences. You can usually get away with this, but if you are _also_ exposing query results directly to the outside world then this could be a breaking change to clients of this api. (If you are having to change the UI layer because the inner api has changed, then something is badly wrong).

chrisfoster76 (changes requested):

src/Approvals/SFA.DAS.Approvals.Api/Controllers/CohortController.cs
alireid: Fixed

src/Approvals/SFA.DAS.Approvals.Api/Controllers/CohortController.cs
alireid: Gotcha, fixed!

src/Approvals/SFA.DAS.Approvals/Application/Cohorts/Queries/GetCohortDetails/GetCohortDetailsQuery.cs
Pauljgraham: Will write a sub type in the response object, but think we might get duplication coverage issues. And I hate the layers of mappings we are introducing. I much prefer the passthrough mechanisms when it's a call just to your own inner API.

chrisfoster76 (approved):

PR #1093, opened 2023-07-03
CSP-783 GetCalendarEvents page and pagesize added

MarkFCain:


src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Models/GetCalendarEventsModel.cs
```diff
@@ -0,0 +1,44 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.ApprenticeAan.Application.CalendarEvents.Queries.GetCalendarEvents;
+using SFA.DAS.ApprenticeAan.Application.Common;
+using SFA.DAS.ApprenticeAan.Application.Infrastructure.Configuration;
+
+namespace SFA.DAS.ApprenticeAan.Api.Models;
+
+public class GetCalendarEventsModel
```
DevenKShah, line 8: Suggest rename to `GetCalendatEventsRequestModel` to differentiate between response and request models

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api/Models/GetCalendarEventsModel.cs
```diff
@@ -0,0 +1,44 @@
...
+    [FromQuery]
+    public List<int> CalendarId { get; set; } = new List<int>();
+
+    [FromQuery]
+    public List<int> RegionId { get; set; } = new List<int>();
+
+    [FromQuery]
+    public int? Page { get; set; }
+
+    [FromQuery] public int? PageSize { get; set; }
```
DevenKShah, line 31: New line after attribute

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application.UnitTests/SFA.DAS.ApprenticeAan.Application.UnitTests.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Shared\SFA.DAS.SharedOuterApi\SFA.DAS.SharedOuterApi.csproj" />
+    <ProjectReference Include="..\SFA.DAS.ApprenticeAan.Api\SFA.DAS.ApprenticeAan.Api.csproj" />
```
DevenKShah, line 28: This is not correct reference, is this needed?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application.UnitTests/CalendarEvents/Queries/GetCalendarEvents/GetCalendarEventsQueryTests.cs
DevenKShah, line 1: This should target the model and not the query, so the tests should be in the API project and not here. This will also allow you to remove the invalid project reference.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application.UnitTests/Services/QueryStringParameterBuilderTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using SFA.DAS.ApprenticeAan.Application.Common;
+using SFA.DAS.ApprenticeAan.Application.Services;
+
+namespace SFA.DAS.ApprenticeAan.Application.UnitTests.Services;
+
+public class QueryStringParameterBuilderTests
+{
+
+    [Test, AutoData]
+    public void Operator_PopulatesModelFromParameters(Guid memberId, DateTime? fromDate, DateTime? toDate,
```
DevenKShah, line 14: Instead of one big test, we need a test per field with both cases where value is null and not null. This will improve your coverage as well.

DevenKShah (commented):
The API project is wrongly reference in Application.UnitTests. I have added hints on how to fix this and also improve coverage

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
```diff
@@ -0,0 +1,227 @@
...
+namespace SFA.DAS.ApprenticeAan.Api.UnitTests.Services;
+
+public class QueryStringParameterBuilderTests
+{
+
+    [Test, AutoData]
+    public void Operator_PopulatesModelFromParameters(Guid memberId, DateTime? fromDate, DateTime? toDate,
+        List<EventFormat> eventFormats, List<int> calendarIds, List<int> regionIds, int? page, int? pageSize)
+    {
+        var model = new GetCalendarEventsRequestModel
```
DevenKShah, line 17: Can you use `sut` for the target object name please

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
```diff
@@ -0,0 +1,227 @@
+﻿using AutoFixture.NUnit3;
+using FluentAssertions;
+using SFA.DAS.ApprenticeAan.Api.Models;
+using SFA.DAS.ApprenticeAan.Application.CalendarEvents.Queries.GetCalendarEvents;
+using SFA.DAS.ApprenticeAan.Application.Common;
+using SFA.DAS.ApprenticeAan.Application.Services;
+
+namespace SFA.DAS.ApprenticeAan.Api.UnitTests.Services;
+
+public class QueryStringParameterBuilderTests
```
DevenKShah, line 10: The target is in Api\Models folder, so this test file needs to be moved to similar folder in the unit test project and it needs to reflect the target name in the filename as well.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        query.FromDate.Should().BeNull();
+        query.ToDate.Should().BeNull();
+        query.Page.Should().BeNull();
+        query.PageSize.Should().BeNull();
+    }
+
+    [Test, AutoData]
+    public void Operator_PopulatesModelFromParametersDatesNull(Guid memberId)
+    {
+        var model = new GetCalendarEventsQuery
```
DevenKShah, line 61: This test is not targeting the correct object. You have created instance of return type instead of source type.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
```diff
@@ -0,0 +1,72 @@
...
+using FluentAssertions;
+using SFA.DAS.ApprenticeAan.Api.Models;
+using SFA.DAS.ApprenticeAan.Application.CalendarEvents.Queries.GetCalendarEvents;
+using SFA.DAS.ApprenticeAan.Application.Common;
+
+namespace SFA.DAS.ApprenticeAan.Api.UnitTests.Controllers.CalendarEventsControllerTests;
+public class GetCalendarEventsRequestModelTests
+{
+    [Test, AutoData]
+    public void Operator_PopulatesModelFromParameters(Guid memberId, DateTime? fromDate, DateTime? toDate, List<EventFormat> eventFormats, List<int> calendarIds, List<int> regionIds, int? page, int? pageSize)
```
DevenKShah, line 11: Why are we testing operator in controller?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        if (calendarId != null)
+        {
+            model.CalendarId = new List<int> { calendarId.Value };
+        }
+
+        var parameters = QueryStringParameterBuilder.BuildQueryStringParameters(model);
+
+        parameters.TryGetValue("calendarId", out var calendarIdResult);
+        if (calendarIdResult != null)
```
DevenKShah, line 147: The calendarId is initialised to empty array so I am not sure when it would ever be null?

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        if (regionId != null)
+        {
+            model.RegionId = new List<int> { regionId.Value };
+        }
+
+        var parameters = QueryStringParameterBuilder.BuildQueryStringParameters(model);
+
+        parameters.TryGetValue("regionId", out var regionIdResult);
+        if (regionIdResult != null)
```
DevenKShah, line 174: Same issue here as calendarId

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
MarkFCain: ok

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
MarkFCain: ok

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
MarkFCain: I have it in the wrong folder, moving it.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
MarkFCain: But the parameter list won't have a match if the calendarId is a list of 0, so the calendarIdResult is null in that case.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Services/QueryStringParameterBuilderTests.cs
MarkFCain: But the parameter list won't have a match if the regionId is a list of 0, so the regionIdResult is null in that case.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
MarkFCain: yes I named it completely wrong!

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Models/GetCalendarEventsQueryTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        query.ToDate.Should().Be(toDate?.ToString("yyyy-MM-dd"));
+        query.EventFormat.Should().BeEquivalentTo(eventFormats);
+        query.Page.Should().Be(page);
+        query.PageSize.Should().Be(pageSize);
+    }
+
+    [Test, AutoData]
+    public void Operator_PopulatesQueryFromModelWithNulls(Guid memberId)
+    {
+        var model = new GetCalendarEventsQuery
```
DevenKShah, line 38: I think this is still not correct as we are not really testing anything here. Ideally I would expect that you populate member id in the model and then do the conversion from model to query and then all your asserts would make sense.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Models/QueryStringParameterBuilderTests.cs
DevenKShah, line 1: This should be in the Service folder as the Static class is in the service folder.

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
DevenKShah: I think this is still not correct as we are not really testing anything here. Ideally I would expect that you populate member id in the model and then do the conversion from model to query and then all your asserts would make sen

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Models/QueryStringParameterBuilderTests.cs
MarkFCain: ok fixed

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Api.UnitTests/Controllers/CalendarEventsControllerTests/GetCalendarEventsQueryTests.cs
MarkFCain: I've added a test that injects a autodata'd model, but not sure if that's what you meant.

DevenKShah (approved):

DevenKShah (approved):

PR #1097, opened 2023-07-10
Mac 188 employer account id type

johrs:


src/ApprenticePortal/SFA.DAS.ApprenticePortal/SFA.DAS.ApprenticePortal.csproj
```diff
@@ -7,6 +7,7 @@
   <ItemGroup>
 	<PackageReference Include="MediatR" />
 	<PackageReference Include="Microsoft.AspNetCore.JsonPatch" />
+	<PackageReference Include="Newtonsoft.Json" />
```
cofaulco, line 10: how come these Newtonsoft references have been added to an unrelated area of the APIM-Endpoints solution?

cofaulco (commented):
1 question raised

src/ApprenticePortal/SFA.DAS.ApprenticePortal/SFA.DAS.ApprenticePortal.csproj
johrs: The solution wouldn't build without them

cofaulco (approved):

PR #1117, opened 2023-08-10
CSP-960 get addresses from locationsApi added, with coverage

MarkFCain:


DevenKShah (changes requested):
This PR is fine, but the config is missing role assignment for location api, DevOps should have not approved it, but they did, so the only way I can stop this from going to test is to request changes here. The change is actually required in the employer config role assignments [here](https://github.com/SkillsFundingAgency/das-employer-config/blob/407fa23468b8d17eab8b5c1e980e780ed31defcb/app-role-assignments/config.json#L512C49-L512C49)

DevenKShah (approved):

PR #1170, opened 2023-10-19
CSP-916 Changed GET to POST to protect PII

DevenKShah:


MarkFCain (approved):

DevenKShah:
This change was deemed unnecessary, instead we will be using telemetry redaction package.

PR #1183, opened 2023-11-01
CSP-1070 Get Calendar Event endpoint added

MarkFCain:


src/AdminAan/SFA.DAS.AdminAan/Application/CalendarEvents/Queries/GetCalendarEvent/GetCalendarEventQueryHandler.cs
DevenKShah, line 1: Do not feel obliged to use Query pattern if there is not much going on, you could use outer api client directly from controller.

DevenKShah (dismissed):

DevenKShah (approved):

PR #1187, opened 2023-11-03
FAI-958 Hot Fix for Email address with +1 AAN Outer APIM

balaji-j:
Removed the Member exist check from LookupMember as this happened in the Inner API methods


dashton82:
No longer required

PR #1197, opened 2023-11-16
CSP-1096 Put CalendarEvent and guests

MarkFCain:


src/AdminAan/SFA.DAS.AdminAan/Application/CalendarEvents/Commands/Update/PutEventCommandHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+            if (command.Guests.Any())
+            {
+                var guestsList = new PutEventGuestsModel(command.Guests);
+                await _apiClient.PutGuestSpeakers(command.CalendarEventId, command.RequestedByMemberId!, guestsList, cancellationToken);
+            }
+
+            return Unit.Value;
+        }
+
+        throw new InvalidOperationException($"An attempt to update a calendarEvent id: {command.CalendarEventId} by member id {command.RequestedByMemberId}, came back with unsuccessful response status: {response.ResponseMessage.StatusCode} with message: {response.StringContent}");
```
DevenKShah, line 31: Two things here, one if you want an exception to be raised on unsuccessful response, just don't use AllowAllStatusCode/Response combo as RestEase client will raise exception for you. 
Secondly, if you want to throw exception manually like this, I would not include the string content in the exception message as that will not be readable at all. You can include this as data if you like.

MarkFCain: Ill remove the AllAllStatusCode and remove the throw

src/AdminAan/SFA.DAS.AdminAan/Application/CalendarEvents/Commands/Update/PutEventCommandHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        _apiClient = apiClient;
+    }
+
+    public async Task<Unit> Handle(PutEventCommand command, CancellationToken cancellationToken)
+    {
+        var response = await _apiClient.PutCalendarEvent(command.RequestedByMemberId!, command.CalendarEventId!,
+            command, cancellationToken);
+
+        if (response.ResponseMessage.StatusCode == System.Net.HttpStatusCode.NoContent && command.Guests.Any())
```
DevenKShah, line 21: You have to assume that the status code is success here as otherwise an exception would have occurred. So no status check is necessary.

DevenKShah (approved):

src/AdminAan/SFA.DAS.AdminAan/Infrastructure/IAanHubRestApiClient.cs
```diff
@@ -28,6 +29,10 @@ public interface IAanHubRestApiClient
     [AllowAnyStatusCode]
     Task<Response<PostEventCommandResult>> PostCalendarEvents([Header(Constants.ApiHeaders.RequestedByMemberIdHeader)] Guid requestedByMemberId, [Body] PostEventCommand command, CancellationToken cancellationToken);
 
+    [Put("calendarEvents/{calendarEventId}")]
+    Task<Response<Unit>> PutCalendarEvent([Header(Constants.ApiHeaders.RequestedByMemberIdHeader)] Guid requestedByMemberId, [Path] Guid calendarEventId, [Body] PutEventCommand command, CancellationToken cancellationToken);
```
DevenKShah, line 33: This should be `Task PutCalendarEvent(....)` so there is no return type as we dont need to inspect the responce, we just let it fail here if it is not successful.

MarkFCain: yes

src/AdminAan/SFA.DAS.AdminAan/Application/CalendarEvents/Commands/Update/PutEventCommandHandler.cs
MarkFCain: ok, removed

DevenKShah (approved):

PR #1211, opened 2023-11-29
Outer api changes for FLP-375

tom-gough:


src/Apprenticeships/SFA.DAS.Apprenticeships.Api/Controllers/ApprenticeshipController.cs
```diff
@@ -1,8 +1,10 @@
-﻿using Microsoft.AspNetCore.Mvc;
+﻿using System.Net;
```
MikeV-TalentConsulting, line 1: Unused using

MikeV-TalentConsulting (approved):

PR #1228, opened 2023-12-11
QF-1678 Fix for not found when querying my apprenticeships

ChrisJWoodcock:


src/ApprenticeFeedback/SFA.DAS.ApprenticeFeedback.Api/Controllers/FeedbackTransactionController.cs
```diff
@@ -30,7 +31,7 @@ public FeedbackTransactionController(IMediator mediator, ILogger<FeedbackTransac
         }
 
         [HttpPost]
-        public async Task<ActionResult<GenerateEmailTransactionResponse>> GenerateEmailTransaction()
+        public async Task<ActionResult<NullResponse>> GenerateEmailTransaction()
```
Jordan-fujitsu, line 34: This might need to be updated depending on your response to my comment on the other PR regarding using `NullResponse `

ChrisJWoodcock: The NullResponse is needed here to follow the pattern used elsewhere - the other Summaries one is not using the pattern correctly but don't want to disturb that too much at this stage

Jordan-fujitsu (approved):

PR #1246, opened 2023-12-22
Fai 961 apim add job categories search results page

balaji-j:
Commit Details:
Adding Job Categories search filter for the Apprenticeship vacancies.

Story: https://skillsfundingagency.atlassian.net/browse/FAI-961


dashton82:
part of #1247

PR #1248, opened 2024-01-03
CSP-1192 Get leaving reasons

MarkFCain:


src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/LeavingReasons/Queries/GetLeavingReasons/GetLeavingReasonsQueryHandler.cs
```diff
@@ -0,0 +1,21 @@
+﻿using MediatR;
+using SFA.DAS.ApprenticeAan.Application.Infrastructure;
+using SFA.DAS.ApprenticeAan.Application.Model;
+
+namespace SFA.DAS.ApprenticeAan.Application.LeavingReasons.Queries.GetLeavingReasons;
+
+public class GetLeavingReasonsQueryHandler : IRequestHandler<GetLeavingReasonsQuery, List<LeavingCategory>>
```
DevenKShah, line 7: I am discouraging use of Handlers if it is as simple as returning the results from inner api. You can just have controller using outer api client directly. Handler should have a purpose, in this case it has none.

DevenKShah (approved):

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/LeavingReasons/Queries/GetLeavingReasons/GetLeavingReasonsQueryResult.cs
DevenKShah, line 1: This is now inner api response class, so it should be Suffixed with Response and not QueryResult. Also needs to be under appropriate namespace.

MarkFCain: Actually this isn't needed, because the nature of the json is just a list, so removing this

DevenKShah (approved):

PR #1253, opened 2024-01-08
CSP-1038 Post member leaving

MarkFCain:


ManishAgarwalDfE (approved):

src/ApprenticeAan/SFA.DAS.ApprenticeAan.Application/Infrastructure/IAanHubRestApiClient.cs
```diff
@@ -95,6 +96,10 @@ public interface IAanHubRestApiClient
         [Body] JsonPatchDocument<PatchMemberRequest> patchMemberRequest,
         CancellationToken cancellationToken);
 
+    [Post("members/{memberId}/Leaving")]
+    [AllowAnyStatusCode]
+    Task<Response<PostMemberLeavingResponse>> PostMembersLeaving([Path] Guid memberId, [Body] PostMemberLeavingModel model, CancellationToken cancellationToken);
```
DevenKShah, line 101: Because you do not have a response body as such, use `Task<HttpResponseMessage>`  instead, this will allow you to get rid of empty response type as well.

MarkFCain: That's good, will do

DevenKShah (approved):

PR #1318, opened 2024-02-12
CSP-1213-Update_RoatpCourseanagement_to_.net8.0

PeterDavidWilliams:


MarkFCain:
We are no longer removing dependencies on shared outer, so this branch is redundant

PR #1339, opened 2024-02-20
Flp 394 send price change request employer initiated

tom-gough:


src/Apprenticeships/SFA.DAS.Apprenticeships.Api/Controllers/ApprenticeshipController.cs
```diff
@@ -68,8 +68,7 @@ public async Task<ActionResult> GetApprenticeshipKey(string apprenticeshipHashed
         {
             await _apiClient.PostWithResponseCode<object>(new PostCreateApprenticeshipPriceChangeRequest(
                 apprenticeshipKey,
-                request.ProviderId,
-                request.EmployerId,
+                request.Requester,
```
MikeV-TalentConsulting, line 71: In the method CreateApprenticeshipPriceChange, an Ok result is returned even if the innerApi fails, I know this is the existing behavior for the provider initiated path, but this could do with returning a negative status code on failure.

tom-gough: sorted

MikeV-TalentConsulting (approved):

PR #1391, opened 2024-03-22
FAI-1106 Select address endpoints

SarahDuncan:
Added (for the 'Select your address' page):

- get endpoint to get addresses by postcode
- post endpoint to post selected address


src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship.Api/Controllers/UsersController.cs
```diff
@@ -85,5 +88,51 @@ public async Task<IActionResult> PostcodeAddress([FromQuery] string postcode)
...
+        [HttpGet]
+        [Route("select-address")]
+        public async Task<IActionResult> SelectAddress([FromQuery] string postcode)
+        {
+            try
+            {
+                var queryResponse = await _mediator.Send(new GetCandidateAddressesByPostcodeQuery(postcode));
+
+                if (queryResponse.AddressesResponse == null || !queryResponse.AddressesResponse.Addresses.Any())
+                    return NotFound();
```
dashton82, line 101: We should really just return an empty list in this case rather than a not found result.

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/GetAddresses/GetAddressesQueryHandler.cs
```diff
@@ -23,9 +24,10 @@ public GetAddressesQueryHandler(ILocationApiClient<LocationApiConfiguration> loc
 
         public async Task<GetAddressesQueryResult> Handle(GetAddressesQuery request, CancellationToken cancellationToken)
         {
-            if (string.IsNullOrEmpty(request.Query)) throw new ArgumentException($"Query is required", nameof(GetAddressesQuery.Query));
+            if (string.IsNullOrEmpty(request.Postcode)) throw new ArgumentException($"Query is required", nameof(GetAddressesQuery.Postcode));
 
-            var addressesResponse = await _locationApiClient.Get<GetAddressesListResponse>(new GetAddressesQueryRequest(request.Query, _config.LocationsApiMinMatch));
+            var postcode = HttpUtility.UrlDecode(request.Postcode);
```
dashton82, line 29: you shouldnt need to do this, when it hits the controller action it will already be done

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/InnerApi/CandidateApi/Requests/PostCandidateAddressApiRequest.cs
```diff
@@ -0,0 +1,27 @@
...
+    private readonly string _govUkIdentifier;
+    public object Data { get; set; }
+
+    public PutCandidateAddressApiRequest(string govIdentifier, PutCandidateAddressApiRequestData data)
+    {
+        _govUkIdentifier = govIdentifier;
+        Data = data;
+    }
+
+    public string PutUrl => $"api/addresses/{_govUkIdentifier}";
```
dashton82, line 15: i think the resource might be better described as:
```suggestion
    public string PutUrl => $"api/candidates/{_candidateId}/addresses/";
```
we should know the candidate id when doing the request

dashton82 (changes requested):

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Commands/Users/Address/CreateAddressCommandHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        var postData = new PutCandidateAddressApiRequestData
+        {
+            Email = request.Email,
+            AddressLine1 = request.AddressLine1,
+            AddressLine2 = request.AddressLine2,
+            AddressLine3 = request.AddressLine3,
+            AddressLine4 = request.AddressLine4,
+            Postcode = request.Postcode
+        };
```
dashton82, line 30: Is it possible to also store the lat/lon at this point as well?

dashton82 (changes requested):

dashton82 (approved):

PR #1401, opened 2024-04-02
FAI-1108 Candidate's notification preferences

SarahDuncan:
Adds a get endpoint to get all preferences and any existing candidate preferences for them. Also adds a post endpoint to save notification preferences.


dashton82:
Included as part of another PR

PR #1411, opened 2024-04-10
QF-1843 Defer apprentice target update when data not available

ChrisJWoodcock:


src/ApprenticeFeedback/SFA.DAS.ApprenticeFeedback.Api/Controllers/ApprenticeController.cs
Jordan-fujitsu, line 1: Instead of removing Exception e is it worth logging it? e.g `_logger.LogError(e, $"Error getting apprentice id {id}");`

ChrisJWoodcock: I'm not going to fix that in this PR

Jordan-fujitsu (approved):

PR #1416, opened 2024-04-16
CSP-1105: Forename & Surname Sync to AAN

shaun-downey-education:
- Sync forename & surname when logging in to AAN for the first time.


shaun-downey-education:
- Abandoned for a branch name change.

PR #1449, opened 2024-04-30
Csp 1375 get legal entities for employer

MarkFCain:


PeterDavidWilliams (approved):

src/EmployerPR/SFA.DAS.EmployerPR/Application/Queries/GetAccountLegalEntities/GetAccountLegalEntitiesQueryHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    private readonly IAccountsApiClient<AccountsConfiguration> _apiClient;
+
+    public GetAccountLegalEntitiesQueryHandler(IAccountsApiClient<AccountsConfiguration> apiClient)
+    {
+        _apiClient = apiClient;
+    }
+    public async Task<GetAccountLegalEntitiesQueryResult> Handle(GetAccountLegalEntitiesQuery request, CancellationToken cancellationToken)
+    {
+        var accountLegalEntities =
```
DevenKShah, line 17: This response should not be nullable, it can be empty though

src/EmployerPR/SFA.DAS.EmployerPR/InnerApi/Requests/GetAccountLegalEntitiesRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+public class GetAccountLegalEntitiesRequest : IGetAllApiRequest
+{
+    private readonly string _hashedAccountId;
+
+    public GetAccountLegalEntitiesRequest(string hashedAccountId)
+    {
+        _hashedAccountId = hashedAccountId;
+    }
+
+    public string GetAllUrl => $"api/accounts/{_hashedAccountId}/legalentities?includeDetails=true";
```
DevenKShah, line 13: Do we not have a request to this endpoint in shared library? if not then we should add this to shared lib

src/EmployerPR/SFA.DAS.EmployerPR/Application/Queries/GetAccountLegalEntities/AccountLegalEntity.cs
DevenKShah, line 1: This should be in shared library (full blown, including all the properties), but check for an existing one first. Also this should be in InnerApi Response namespace GetAccountLegalEntitiesResponse. But do check existing bits before jumping to this suggestion.

MarkFCain: moved

src/EmployerPR/SFA.DAS.EmployerPR/Application/Queries/GetAccountLegalEntities/GetAccountLegalEntitiesQueryHandler.cs
MarkFCain: fixed

src/EmployerPR/SFA.DAS.EmployerPR/Application/Queries/GetAccountLegalEntities/AccountLegalEntity.cs
MarkFCain: ok done

PeterDavidWilliams (approved):

PR #1464, opened 2024-05-10
Csp 1380 get permissions

MarkFCain:


src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPermissionsRequest.cs
```diff
@@ -0,0 +1,25 @@
...
+    public string? PublicHashedId { get; set; }
+
+    public GetPermissionsRequest(long? ukprn, string? publicHashedId)
+    {
+        Ukprn = ukprn;
+        PublicHashedId = publicHashedId;
+    }
+}
+
+public enum Operation : short
```
DevenKShah, line 20: This enum is used in Response, but is hidden in Request object file. 

Also I think we should have a folder within Requests and Response to separate objects out which related to ProviderRelationships

src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPermissionsRequest.cs
DevenKShah, line 1: This should be under ProviderRelationships folder

src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPermissionsRequest.cs
MarkFCain: moved to own class in repsonses

src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/GetPermissionsRequest.cs
MarkFCain: fixed

DevenKShah (approved):

PR #1476, opened 2024-05-22
Csp 1390 add call to get account provider legal entities

PeterDavidWilliams:


src/Shared/SFA.DAS.SharedOuterApi/Models/Operation.cs
DevenKShah, line 1: There is already an Operation enum defined in `SFA.DAS.SharedOuterApi.InnerApi.Requests.ProviderPermissions.Operation` namespace, it also has a request object, may be just take the request object out from there and into its own file.

src/Shared/SFA.DAS.SharedOuterApi/InnerApi/Requests/ProviderPermissions/GetProviderAccountLegalEntitiesRequest.cs
```diff
@@ -1,15 +1,26 @@
...
     {
         _ukprn = ukprn;
+        if (!operations.IsNullOrEmpty())
+        {
+            foreach (Operation operation in operations)
+                this._operations += $"&operations={(int)operation}";
+        }
+        else
+            _operations = "&operations=";
     }
```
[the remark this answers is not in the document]
DevenKShah: Also by defaulting operations parameter, you could have avoided all the failures in other areas. Thinking about it there was no need to add the operations to the request at all as it would have worked with out it, but I glad you did, we can use it elsewhere.

src/Shared/SFA.DAS.SharedOuterApi/Services/AccountLegalEntityPermissionService.cs
```diff
@@ -26,17 +27,17 @@ public async Task<AccountLegalEntityItem> GetAccountLegalEntity(AccountIdentifie
                 case AccountType.Provider:
                     var providerResponse =
                         await _providerRelationshipsApiClient.Get<GetProviderAccountLegalEntitiesResponse>(
-                            new GetProviderAccountLegalEntitiesRequest(accountIdentifier.Ukprn));
-                    
+                            new GetProviderAccountLegalEntitiesRequest(accountIdentifier.Ukprn, new List<Operation>()));
```
DevenKShah, line 30: If you default it in the ctor, you need not make this change :-)

DevenKShah (approved):

DevenKShah (dismissed):

DevenKShah (approved):

DevenKShah (approved):

PR #1483, opened 2024-05-29
CSP-1379: Get Employer Relationships - Outer API endpoint.

shaun-downey-education:
- Added a **EmployerRelationshipsController** to the **EmployerPR** API as a route through to the **SFA.DAS.PR.Api**


PeterDavidWilliams (approved):

src/Shared/SFA.DAS.SharedOuterApi/Models/Operation.cs
DevenKShah, line 1: Why was this moved out of our namespace?

DevenKShah (changes requested):
Build has failed as the Operation enum has been moved

src/Shared/SFA.DAS.SharedOuterApi/Models/Operation.cs
shaun-downey-education: @DevenKShah The Operation class was duplicated in a few different places so I put it in Models as it is technically an entity linked model.

PR #1486, opened 2024-05-29
CSP-1387 Created solution for ProviderPR outer api

DevenKShah:


src/ProviderPR/SFA.DAS.ProviderPR.Api/SFA.DAS.ProviderPR.Api.http
```diff
@@ -0,0 +1,6 @@
+@SFA.DAS.ProviderPR.Api_HostAddress = http://localhost:5119
```
MarkFCain, line 1: What does this file do?

MarkFCain (approved):

HussainsGitHub (approved):

PR #1493, opened 2024-06-04
FAI-1266-Login-To-Your-Old-Account

chrisfoster76:


src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/CreateAccount/SignIntoYourOldAccount/GetSignIntoYourOldAccountQueryHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+
+public class GetSignIntoYourOldAccountQueryHandler(
+    IFindApprenticeshipLegacyApiClient<FindApprenticeshipLegacyApiConfiguration> legacyApiClient)
+    : IRequestHandler<GetSignIntoYourOldAccountQuery, GetSignIntoYourOldAccountQueryResult>
+{
+    public async Task<GetSignIntoYourOldAccountQueryResult> Handle(GetSignIntoYourOldAccountQuery request, CancellationToken cancellationToken)
+    {
+        var result =
+            await legacyApiClient.Get<GetLegacyValidateCredentialsApiResponse>(
+                new GetLegacyValidateCredentialsApiRequest(request.Email, request.Password));
```
dashton82, line 20: I _think_ in here we should also store a record of a unsuccessful response, we can then check this cache first and see if there is more than 5 failures to just return false.

dashton82 (changes requested):

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/CreateAccount/SignIntoYourOldAccount/GetSignIntoYourOldAccountQueryHandler.cs
dashton82: I think also before we call this we should call the candidate account api with the email address and make sure that it hasnt already been migrated by another user.

dashton82 (approved):

PR #1511, opened 2024-06-14
FAI-1132-Settings-Page

chrisfoster76:


src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship.Api/Startup.cs
```diff
@@ -90,7 +90,7 @@ public void ConfigureServices(IServiceCollection services)
             }
             
             services.AddApplicationInsightsTelemetry(_configuration["APPINSIGHTS_INSTRUMENTATIONKEY"]);
-            services.AddOpenTelemetryRegistration(_configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"]);
+            //services.AddOpenTelemetryRegistration(_configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"]);
```
dashton82, line 93: I think this needs to be uncommeted.
I reckon its worth while adding a stub implementation of the IMetrics so we have something like
```
    public class StubFindAnApprenticeshipMetrics : IMetrics
    {
        public void IncreaseVacancyViews(string vacancyReference, int viewCount = 1)
        {
        }
    }
```
and the extension is changed to:
```
public static class AddOpenTelemetryExtensions
    {
        /// <summary>
        /// Add the OpenTelemetry telemetry service to the application.
        /// </summary>
        /// <param name="services">Service Collection</param>
        /// <param name="appInsightsConnectionString">Azure app insights connection string.</param>
        public static void AddOpenTelemetryRegistration(this IServiceCollection services, string appInsightsConnectionString)
        {
            if (!string.IsNullOrEmpty(appInsightsConnectionString))
            {
                // This service will collect and send telemetry data to Azure Monitor.
                services.AddOpenTelemetry().UseAzureMonitor(options =>
                    {
                        options.ConnectionString = appInsightsConnectionString;
                    })
                    // Configure metrics
                    .WithMetrics(opts => opts
                        .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService(
                            Constants.OpenTelemetry.ServiceName,
                            nameof(FindAnApprenticeship)))
                        .AddMeter(Constants.OpenTelemetry.ServiceMeterName));
                services.AddSingleton<IMetrics, FindAnApprenticeshipMetrics>();
            }
            else
            {
                services.AddSingleton<IMetrics, StubFindAnApprenticeshipMetrics>();
            }
        }
    }
    ```

File: src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/Apply/GetApplicationSubmitted/GetApplicationSubmittedQueryHandler.cs [hunk 2]
```diff
@@ -26,7 +25,14 @@ public class GetApplicationSubmittedQueryHandler : IRequestHandler<GetApplicatio
 
     public async Task<GetApplicationSubmittedQueryResult> Handle(GetApplicationSubmittedQuery request, CancellationToken cancellationToken)
     {
-        var application = await _candidateApiClient.Get<GetApplicationApiResponse>(new GetApplicationApiRequest(request.CandidateId, request.ApplicationId, false));
+        var applicationTask = _candidateApiClient.Get<GetApplicationApiResponse>(new GetApplicationApiRequest(request.CandidateId, request.ApplicationId, false));
+        var aboutYouTask = _candidateApiClient.Get<GetAboutYouItemApiResponse>(new GetAboutYouItemApiRequest(request.CandidateId));
```
dashton82, line 29: I _think_ this is probably better than setting the flag on the first request to true and bringing back the whole application 🤔

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/CreateAccount/SignIntoYourOldAccount/GetSignIntoYourOldAccountQueryHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            cacheItem.SignInAttempts.RemoveAll(x => x < dateTimeService.UtcNow.AddMinutes(-30));
+
+            if (cacheItem.SignInAttempts.Count >= 5)
+            {
+                return new GetSignIntoYourOldAccountQueryResult { IsValid = false };
+            }
+        }
+
+        var result =
```
dashton82, line 35: I think as well as doing this we also want to check the email hasnt already been migrated? unless this is done somewhere else

PR #1516, opened 2024-06-18
CSP-1411 - Post Permissions Endpoint.

shaun-downey-education:
- New endpoint as a pass through for the PostPermissions endpoint on CSP-1392.

### DIFF
--- a/src/EmployerPR/SFA.DAS.EmployerPR.Api/Controllers/PermissionsController.cs
+++ b/src/EmployerPR/SFA.DAS.EmployerPR.Api/Controllers/PermissionsController.cs
@@ -1,5 +1,6 @@
 ﻿using MediatR;
 using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.EmployerPR.Application.Commands.PostPermissions;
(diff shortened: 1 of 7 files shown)


src/EmployerPR/SFA.DAS.EmployerPR.Api/Controllers/PermissionsController.cs
```diff
@@ -19,4 +20,15 @@ public async Task<IActionResult> GetPermissions([FromQuery] GetPermissionsQuery
 
         return Ok(result);
     }
+
+    [HttpPost]
+    [Produces("application/json")]
+    [ProducesResponseType(typeof(PostPermissionsCommandResult), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
DevenKShah, line 27: Bad request is not being produced. Usually all the validations are happening in the FE, so these requests are not expected to fail validations in the inner, so we do not care about bad request, if it fails any validation in the inner then we just let the error bubble up.

DevenKShah (dismissed):
Just need to remove bad request response type from the controller

PR #1520, opened 2024-06-20
CSP-1367 - Updated PR API call request url and to not hard code the operations list

PeterDavidWilliams:


src/Reservations/SFA.DAS.Reservations/Application/ProviderAccounts/Queries/GetProviderAccountLegalEntitiesWithCreatCohortQueryHandler.cs
```diff
@@ -12,6 +14,6 @@ public class GetProviderAccountLegalEntitiesWithCreatCohortQueryHandler(IProvide
 {
     public async Task<GetProviderAccountLegalEntitiesWithCreateCohortResult> Handle(GetProviderAccountLegalEntitiesWithCreatCohortQuery request, CancellationToken cancellationToken)
     {
-        return await apiClient.Get<GetProviderAccountLegalEntitiesWithCreateCohortResult>(new GetProviderAccountLegalEntitiesWithCreatCohortRequest(request.Ukprn));
+        return await apiClient.Get<GetProviderAccountLegalEntitiesWithCreateCohortResult>(new GetProviderAccountLegalEntitiesWithCreatCohortRequest(request.Ukprn, new List<Operation> { Operation.CreateCohort }));
```
DevenKShah, line 17: The request is specific to Cohort, we should get rid of this and use the generic one that you have used else where

PR #1538, opened 2024-06-29
Aad 283 contentful integration

alireid:


src/ApprenticeApp/SFA.DAS.ApprenticeApp/Application/Commands/SupportAndGuidance/AddUpdateApprenticeArticleCommandHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class AddUpdateApprenticeArticleCommandHandler : IRequestHandler<AddUpdateApprenticeArticleCommand, Unit>
+    {
+        private readonly ILogger<AddUpdateApprenticeArticleCommandHandler> _logger;
+        private readonly IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration> _accountsApiClient;
+
+        public AddUpdateApprenticeArticleCommandHandler
+        (
+            ILogger<AddUpdateApprenticeArticleCommandHandler> logger,
+            IApprenticeAccountsApiClient<ApprenticeAccountsApiConfiguration> accountsApiClient
+
```
CathyGroom, line 21: whitespace

src/ApprenticeApp/SFA.DAS.ApprenticeApp/Application/Commands/SupportAndGuidance/AddUpdateApprenticeArticleCommandHandler.cs
CathyGroom, line 1: Line 21: whitespace

src/ApprenticeApp/SFA.DAS.ApprenticeApp/Services/ContentService.cs
CathyGroom, line 1: Do we need logging if the get by id methods return null?

src/ApprenticeApp/SFA.DAS.ApprenticeApp.Api/Controllers/SupportAndGuidanceController.cs
CathyGroom, line 1: Line 55: if the query result is added to the return method it resolves the warning on line 50

CathyGroom (changes requested):
All looks really good.  Just commented on a couple of issues.

src/ApprenticeApp/SFA.DAS.ApprenticeApp/Services/ContentService.cs
alireid: I think so.. just so that we know the method has been hit, in a case where this has been deployed and we can't step into code all we would have is the logs and I believe it may be handy in a circumstance to follow through the stack to determine where problems may be.

src/ApprenticeApp/SFA.DAS.ApprenticeApp.Api/Controllers/SupportAndGuidanceController.cs
alireid: Ah yes, handy to return the result. I dont think we are using GetOneByEntryId as at now, but I surfaced it in case we ever do. Added that in though.

CathyGroom (approved):
Approved.

CathyGroom (approved):
Approved.

CathyGroom (approved):
Approved.

PR #1556, opened 2024-07-08
CSP-1575 POST permissions removal added

MarkFCain:


src/EmployerPR/SFA.DAS.EmployerPR/Infrastructure/IProviderRelationshipsApiRestClient.cs
```diff
@@ -15,4 +15,7 @@ public interface IProviderRelationshipsApiRestClient
 
     [Post("permissions")]
     Task<PostPermissionsCommandResult> PostPermissions([Body] PostPermissionsCommand command, CancellationToken cancellationToken);
```
DevenKShah, line 17: There should be no response coming back from a POST, so I do not get this

src/EmployerPR/SFA.DAS.EmployerPR/Application/Commands/PostPermissions/PostPermissionsCommandHandler.cs
```diff
@@ -5,8 +5,15 @@ namespace SFA.DAS.EmployerPR.Application.Commands.PostPermissions;
 
 public class PostPermissionsCommandHandler(IProviderRelationshipsApiRestClient _providerRelationshipsApiRestClient) : IRequestHandler<PostPermissionsCommand, PostPermissionsCommandResult>
```
DevenKShah, line 6: Should return Unit here.

DevenKShah (dismissed):

MarkFCain:
deleted as work done in branch CSP-1473

PR #1571, opened 2024-07-15
CSP-1473: PR Employer Outer API: Send Email to Provider on Permission Created/Updated/Removed

shaun-downey-education:
- Extend the post permissions endpoint to handle notification creation for create, update and removing of permissions.


shaun-downey-education:
Requires a rebase.

PR #1589, opened 2024-07-25
CSP-1433 GetRelationshipByEmail extension

MarkFCain:


src/ProviderPR/SFA.DAS.ProviderPR/Application/Queries/GetRelationshipByEmail/GetRelationshipByEmailQueryHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        var hasUserAccount = false;
+        bool? hasOneEmployerAccount = null;
+        bool? hasOneLegalEntity = null;
+
+        var result = await accountsApiClient.Get<GetUserByEmailResponse>(new GetUserByEmailRequest(request.Email));
+
+        if (result == null)
+        {
+            return new GetRelationshipByEmailQueryResult(hasUserAccount, hasOneEmployerAccount, null, hasOneLegalEntity);
```
shaun-downey-education, line 27: Would suggest using either non-mandatory properties or class based defaults instead of using method based properties as setters, reduces the number of times the properties hasOneLegalEntity and hasOneEmployerAccount have to be referenced before they are actually required for use (would allow them to be completely removed even).

Would allow GetRelationshipByEmailQueryResult(hasUserAccount, hasOneEmployerAccount, null, hasOneLegalEntity) to be reduced to GetRelationshipByEmailQueryResult() when the result is null.

src/ProviderPR/SFA.DAS.ProviderPR/Application/Queries/GetRelationshipByEmail/GetRelationshipByEmailQueryResult.cs
```diff
@@ -0,0 +1,40 @@
...
+    public bool? HasOneEmployerAccount { get; init; }
+    public long? AccountId { get; init; }
+    public bool? HasOneLegalEntity { get; init; }
+    public string? AccountLegalEntityPublicHashedId { get; init; }
+    public long? AccountLegalEntityId { get; init; }
+    public string? AccountLegalEntityName { get; init; }
+    public bool? HasRelationship { get; init; }
+    public List<Operation>? Operations { get; init; } = new List<Operation>();
+
+    public GetRelationshipByEmailQueryResult(bool hasUserAccount, bool? hasOneEmployerAccount, long? accountId,
```
shaun-downey-education, line 17: As suggested in the previous comment, if the properties are known to be defaulted to null then I would suggest having them set as null on the constructor allowing for the creation of the object without the need to set them explicitly.

src/ProviderPR/SFA.DAS.ProviderPR/Application/Queries/GetRelationship/GetRelationshipResponse.cs
DevenKShah, line 1: There is a namespace for inner api request and response objects, this should be moved in that space

src/ProviderPR/SFA.DAS.ProviderPR/Application/Queries/GetRelationship/PermissionAction.cs
DevenKShah, line 1: There is a namespace for inner api request and response objects, this should be moved in that space

src/ProviderPR/SFA.DAS.ProviderPR/Application/Queries/GetRelationshipByEmail/GetRelationshipByEmailQueryHandler.cs
DevenKShah: I agree. I would have done it slightly differently as well. Since this is an incremental population of result object and we may not need to populate other bits as we have exit route at each step, I would instantiate the result object right at the top (as this needs to be returned no matter what) and then as I go I would collect information directly into the result object and return it as I hit an exit route. It will require all the properties in the result object to have default values or be nullable. And we can see exactly where the exit conditions are. In my opinion it will read lot better.

MarkFCain:
need to rebase on changed root means this is now in a different branch

PR #1607, opened 2024-08-09
Qf-2052-Check_amswers_apim

karen-hanna:


src/ProviderRequestApprenticeTraining/SFA.DAS.ProviderRequestApprenticeTraining.Api/Models/EmployerRequests.cs
```diff
@@ -0,0 +1,63 @@
...
+        public int StandardLevel { get; set; }
+        public List<EmployerRequest> Requests { get; set; }
+
+        public static implicit operator EmployerRequests(GetEmployerRequestsByIdsResult source)
+        {
+            if (source.EmployerRequests.Any())
+            {
+                return new EmployerRequests
+                {
+                    StandardReference = source.EmployerRequests.FirstOrDefault().StandardReference,
```
ChrisJWoodcock, line 23: This was a Sonar cloud warning in another place - it should be First() as you have already done Any()

ChrisJWoodcock (changes requested):

src/ProviderRequestApprenticeTraining/SFA.DAS.ProviderRequestApprenticeTraining.Api/Models/EmployerRequests.cs
karen-hanna: Also updated in SelectEmployerRequests as I had done the same there

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

PR #1613, opened 2024-08-13
CSP-1470: Provider PR Outer API: Lookup Request by Id

shaun-downey-education:
- Added **GetRequest** method to the **RequestsController** and added associated unit testing.


MarkFCain (approved):

shaun-downey-education:
Parent pull request CSP-1468 incorporates both 1470 & 1425 as well as 1468.

PR #1662, opened 2024-09-13
CSP-1533: Declined Permissions Request

shaun-downey-education:
- Added new **RequestsController** endpoint called **DeclinedRequest** to handle the decline of a permissions request.

* Restructured application layer to match existing projects structures.


src/EmployerPR/SFA.DAS.EmployerPR.Api/Controllers/RequestsController.cs
```diff
@@ -22,4 +24,20 @@ public async Task<IActionResult> GetRequest([FromRoute] Guid requestId, Cancella
 
         return Ok(result);
     }
+
+    [HttpPost("{requestId:guid}/permission/declined")]
+    [ProducesResponseType(typeof(GetRequestResponse), StatusCodes.Status200OK)]
```
DevenKShah, line 29: Remove return type, we are not returning anything

DevenKShah (approved):

PR #1676, opened 2024-09-18
CSP-1524: Employer PR Outer API: Accept Add Account & Permissions Request

shaun-downey-education:
- New **RequestsController** **AcceptAddAccountRequest** with associated **AcceptAddAccountRequestHandler** and associated unit testing.


src/EmployerPR/SFA.DAS.EmployerPR.Api/Controllers/RequestsController.cs
```diff
@@ -28,7 +29,7 @@ public async Task<IActionResult> GetRequest([FromRoute] Guid requestId, Cancella
     }
 
     [HttpPost("{requestId:guid}/permission/declined")]
-    [ProducesResponseType(typeof(GetRequestResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(Unit), StatusCodes.Status200OK)]
```
DevenKShah, line 32: We do not want to export Mediator's Unit from our controller, type is optional param, we can just remove it.

DevenKShah (dismissed):

PR #1760, opened 2024-11-05
FAI-1954 - Add search alert to no results page

killij:


killij:
Includes changes for 1945 as well.

PR #1772, opened 2024-11-11
FAI-2051 Delete Saved Searches On Account Deletion

balaji-j:
Functionality added to Delete the Saved Searches when an account gets deleted
Added missing unit test cases
Minor code clean-up

Story: https://skillsfundingagency.atlassian.net/browse/FAI-2051


src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Commands/SaveSearch/CreateSaveSearch/SaveSearchCommand.cs
```diff
@@ -2,16 +2,16 @@
...
 public record SaveSearchCommand(
     Guid Id,
     Guid CandidateId,
     bool DisabilityConfident,
     decimal? Distance,
-    string? Location,
-    string? SearchTerm,
-    List<int>? SelectedLevelIds,
-    List<int>? SelectedRouteIds,
+    string Location,
```
killij, line 12: Is there a reason to make these non-nullable? This could introduce bugs unless every representation of the search parameters is non-nullable.

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Commands/Users/DeleteCandidate/DeleteCandidateCommandHandler.cs
```diff
@@ -18,6 +18,7 @@
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
+using SFA.DAS.FindAnApprenticeship.InnerApi.FindApprenticeApi.Requests;
 
 namespace SFA.DAS.FindAnApprenticeship.Application.Commands.Users.DeleteCandidate
 {
```
killij, line 24: Not an issue with this PR but the handler shouldn't be a record, it should be a class.

balaji-j: Replaced to class instead of record type

PR #1778, opened 2024-11-11
Aad 690 optel integration new

tdavande:
unit tests added


dashton82 (changes requested):
There is the option to add Telemetry with metrics in the shared project. This will then align it with use of the Business Metrics API.

alireid (approved):

alireid (changes requested):
Two tests needs fixing to pass

alireid (approved):
approved in principle, await test

dashton82 (approved):

PR #1789, opened 2024-11-14
remove provider ID from auto create reservation as this determines it as a provider created reservation.

cofaulco:


src/Approvals/SFA.DAS.Approvals/Application/Cohorts/Commands/CreateCohort/CreateCohortCommandHandler.cs
```diff
@@ -39,7 +39,6 @@ public async Task<CreateCohortResult> Handle(CreateCohortCommand request, Cancel
                     AccountId = request.AccountId,
                     AccountLegalEntityId = request.AccountLegalEntityId,
                     CourseCode = request.CourseCode,
-                    ProviderId = request.ProviderId,
```
Pauljgraham, line 42: I can see ProviderId is optional on the Reservations side, but there is a uint cast on the following line which may cause this to fail or pass a zero through, not sure what would happen if it's zero

https://github.com/SkillsFundingAgency/das-apim-endpoints/blob/master/src/Approvals/SFA.DAS.Approvals/Services/AutoReservationsService.cs#L63

src/Approvals/SFA.DAS.Approvals/Application/DraftApprenticeships/Commands/AddDraftApprenticeship/AddDraftApprenticeshipCommandHandler.cs
```diff
@@ -39,7 +39,6 @@ public async Task<AddDraftApprenticeshipResult> Handle(AddDraftApprenticeshipCom
                     AccountId = cohort.AccountId,
                     AccountLegalEntityId = cohort.AccountLegalEntityId,
                     CourseCode = request.CourseCode,
-                    ProviderId = cohort.ProviderId,
```
Pauljgraham, line 42: same concern as above

Pauljgraham (approved):

PR #1791, opened 2024-11-15
FAI-1922 APIM Enable sort by salary with Competitive salary included

balaji-j:
https://skillsfundingagency.atlassian.net/browse/FAI-1922


src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/SearchApprenticeships/SearchApprenticeshipsQueryHandler.cs
```diff
@@ -62,6 +63,14 @@ public async Task<SearchApprenticeshipsResult> Handle(SearchApprenticeshipsQuery
             var categories = routes.Routes.Where(route => request.SelectedRouteIds != null && request.SelectedRouteIds.Contains(route.Id.ToString()))
                 .Select(route => route.Name).ToList();
 
+            var totalWageTypeVacanciesCount = new GetApprenticeshipCountResponse { TotalVacancies = 0 };
```
killij, line 66: Could this run in parallel with the api call to get the vacancies?

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/InnerApi/Requests/GetApprenticeshipCountRequest.cs
```diff
@@ -1,10 +1,11 @@
+using SFA.DAS.FindAnApprenticeship.Domain.Models;
 using SFA.DAS.SharedOuterApi.Interfaces;
 
 namespace SFA.DAS.FindAnApprenticeship.InnerApi.Requests
 {
-    public class GetApprenticeshipCountRequest : IGetApiRequest
+    public class GetApprenticeshipCountRequest(WageType? wageType = null) : IGetApiRequest
     {
-        public string GetUrl => "/api/vacancies/count";
+        public string GetUrl => $"/api/vacancies/count?wageType={wageType}";
```
killij, line 8: Careful adding null values into query string parameters, they can sometimes get interpreted in unpredictable ways depending on the type.

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/InnerApi/Requests/GetVacanciesRequest.cs
```diff
@@ -4,46 +4,26 @@
...
-            _disabilityConfident = disabilityConfident;
-
-        }
+        private readonly string _categories = categories is { Count: > 0 } ? string.Join("&categories=", categories) : string.Empty;
+        private readonly string _levels = levels is { Count: > 0 } ? string.Join("&levels=", levels) : string.Empty;
 
 
         public string Version => "2.0";
-        public string GetUrl => $"/api/vacancies?lat={_lat}&lon={_lon}&distanceInMiles={_distance}&sort={_sort}&pageNumber={_pageNumber}&pageSize={_pageSize}&categories={_categories}&levels={_levels}&searchTerm={_searchTerm}&disabilityConfident={_disabilityConfident}";
+        public string GetUrl =>
```
killij, line 26: Perfect example of potential problems with empty types. You might find on the other end of this that categories gets created as an empty list with a single null item in it, if it is passed as an empty string.

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/Application/Queries/SearchApprenticeships/SearchApprenticeshipsQueryHandler.cs
balaji-j: Because this call is required when the sort type is by SalaryAsc or SalaryDesc, so it would be good to keep it conditional

src/FindAnApprenticeship/SFA.DAS.FindAnApprenticeship/InnerApi/Requests/GetApprenticeshipCountRequest.cs
balaji-j: Used StringBuilder to construct the Url, but we have to come up with a QueryBuilderExtension class to handle it through out the application since we heavy use this kind of Url builder all over the place.

dashton82 (approved):

dashton82 (approved):

PR #1801, opened 2024-11-25
FAI-1955 - Correct type for course level

dashton82:


killij (approved):

dashton82:
This is no longer required

PR #1933, opened 2025-02-10
Flp 696 fm36 block updates for private beta withdrawals

MikeV-TalentConsulting:


chrisfoster76 (dismissed):

src/Shared/SFA.DAS.SharedOuterApi/Common/Constants.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace SFA.DAS.SharedOuterApi.Common
+{
+    public static class Constants
```
tom-gough, line 3: What is the thinking behind moving this to here? Maybe quick chat?

tom-gough (changes requested):

tom-gough (approved):

PR #1951, opened 2025-02-17
FAI-2265 ML notification emails

balaji-j:
Update tests for vacancy and application handling

- Excluded `Country` from equivalence checks in `WhenMappingMediatrResponseToGetApprenticeshipVacancyApiResponse`.
- Added `postcode` and `employerLocationOption` parameters in `WhenHandlingSubmitApplicationCommand` and updated related test cases.
- Modified `WhenHandlingWithdrawApplicationCommand` to include new fields and added tests for multiple locations.
- Updated `WhenHandlingDeleteCandidateCommand` to handle new fields and verify deletion of saved searches and accounts.

- Adjusted `WhenHandlingProcessApplicationReminder` and `WhenHandlingProcessVacancyClosedEarlyCommand` to include `employerLocationOption` and removed redundant tests.
- Introduced `WhenGettingWorkLocation` class to test `GetVacancyWorkLocation` method with various location scenarios.


src/Shared/SFA.DAS.SharedOuterApi/Extensions/EmailTemplateAddressExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (address is null) return string.Empty;
+            var city = address.GetLastNonEmptyField();
+            return string.IsNullOrWhiteSpace(city) ? address.Postcode! : $"{city} ({address.Postcode})";
+        }
+
+        public static string GetEmploymentLocationCityNames(List<Address> addresses)
+        {
+            var cityNames = addresses
+                .Select(address => address.GetLastNonEmptyField())
+                .OfType<string>()
```
killij, line 29: The OfType<string> is not required here.

Also, does this method handle the scenario where you have multiple different address in a single city? As I think the format in that scenario is "City (N locations available)"

killij (approved):

src/Shared/SFA.DAS.SharedOuterApi/Extensions/EmailTemplateAddressExtension.cs
balaji-j: Addressed the PR comments

PR #1972, opened 2025-02-21
FAI-2350 FAI-2328 FAI-2329 - Vacancy Closed Handler error

dashton82:
Fix issue where handler was querying the live vacancy view type rather than the closed. I think this was done previously as the function app was receiving the event before the vacancy had been moved to the closed vacancy projection. Added polly to back off and retry then if not found throw an exception so it will be reprocessed by the calling function


killij (dismissed):

balaji-j (approved):

dashton82:
Messed up with name/case changing in adop

PR #2044, opened 2025-03-12
Faisal.ashraf/outer api change qualifications

FaisalAshraf-dfe:
outer api changes for change qualifications with paging.


src/ApprenticeApp/SFA.DAS.ApprenticeApp.Api/Controllers/TasksController.cs
robryb-dfe, line 1: I get a bit nervous when changes like these pop up for other teams, we might be overwriting changes they've made or causing issues

FaisalAshraf-dfe: pulled latest from sprint 5 .

robryb-dfe (approved):

PR #2076, opened 2025-03-19
Award 575 delete form functionality

HamzaIbraheem-DFE:


src/Aodp/SFA.DAS.Aodp.Api/Controllers/PingController.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [ApiController]
+    [Route("[controller]")]
+    public class PingController : Controller
+    {
+        [AllowAnonymous]
+        [HttpGet("/Ping")]
+        public IActionResult Ping()
+        {
+            return Ok("Pong");
```
robryb-dfe, line 14: Should the outer api ping the inner api?

robryb-dfe (approved):

src/Aodp/SFA.DAS.Aodp.Api/Controllers/PingController.cs
HamzaIbraheem-DFE: Probably, as a health check. This is just me moving the controller into a more suitable area, think we need something on the backlog for proper health checks

PR #2123, opened 2025-04-02
Award 669 survey outerapi

raviudari01:


src/Aodp/SFA.DAS.Aodp/Application/Commands/Application/Application/SaveSurveyCommandHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+        public async Task<BaseMediatrResponse<EmptyResponse>> Handle(SaveSurveyCommand request, CancellationToken cancellationToken)
+        {
+            var response = new BaseMediatrResponse<EmptyResponse>()
+            {
+                Success = false
+            };
+
+            try
+            {
+                var result = await _apiClient.PostWithResponseCode<EmptyResponse>(new SaveSurveyApiRequest()
```
HamzaIbraheem-DFE, line 26: If using PostWithResponseCode then need to call result.EnsureSuccess after

HamzaIbraheem-DFE (approved):

PR #2159, opened 2025-04-17
CSP-1692: Fixed double encoding in training provider details endpoint request

shaun-downey-education:
- Removed location encoding to allow for the rematching of location to the related shortlist item.


src/FindApprenticeshipTraining/SFA.DAS.FindApprenticeshipTraining/InnerApi/Requests/GetCourseProviderDetailsRequest.cs
```diff
@@ -31,7 +31,7 @@ private string BuildUrl()
 
         if (!string.IsNullOrWhiteSpace(Location))
         {
-            query["location"] = HttpUtility.UrlEncode(Location);
+            query["location"] = Location;
```
DevenKShah, line 34: This can have spaces, why are we removing it?

DevenKShah (approved):

PR #2277, opened 2025-07-09
Csp 2077 delete location endpoint

MarkFCain:


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.Api/Controllers/ProviderLocationDeleteController.cs
```diff
@@ -0,0 +1,30 @@
...
+using SFA.DAS.RoatpCourseManagement.Application.Locations.Commands.DeleteProviderLocation;
+
+namespace SFA.DAS.RoatpCourseManagement.Api.Controllers;
+
+[ApiController]
+[Route("providers/{ukprn}/locations/{id}")]
+public class ProviderLocationDeleteController(ILogger<ProviderLocationDeleteController> logger, IMediator mediator) : ControllerBase
+{
+    [HttpDelete]
+    public async Task<IActionResult> DeleteProviderLocation([FromRoute] int ukprn, [FromRoute] Guid id, DeleteProviderLocationCommand command)
```
DevenKShah, line 16: Adding command here gives wrong impression of adding ukprn and location id in to the body. Use a model instead with the properties that you need and then create a command in the action.

MarkFCain: done

PR #2315, opened 2025-08-05
FLP-1201 Update Learning Support

MikeV-TalentConsulting:


src/LearnerData/SFA.DAS.LearnerData/Application/UpdateLearner/EarningsApiClientExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+using SFA.DAS.SharedOuterApi.Configuration;
+using SFA.DAS.SharedOuterApi.InnerApi.Requests.LearnerData;
+using SFA.DAS.SharedOuterApi.Interfaces;
+
+namespace SFA.DAS.LearnerData.Application.UpdateLearner;
+
+/// <summary>
+/// Extensions for the Earnings API client to handle updates related to learners. This is specifically for use with the UpdateLearnerCommandHandler.
+/// </summary>
+internal static class EarningsApiClientExtensions
```
tom-gough, line 11: I like this approach for handling the earnings api calls a lot. Passing the command and letting this class work out what to pass to the earnings endpoints is clean, but it also means if we switch it up to a single endpoint in future this will be the only class that has to change.

tom-gough (approved):

PR #2359, opened 2025-09-03
FAI-2702 Create new provider notification event

balaji-j:
New endpoint added to handle the SharedApplicationReviewedEvent Handled

### DIFF
--- a/src/Recruit/SFA.DAS.Recruit/Handlers/SharedApplicationReviewedEventHandler.cs
+++ b/src/Recruit/SFA.DAS.Recruit/Handlers/SharedApplicationReviewedEventHandler.cs
@@ -1,0 +1,65 @@
+﻿using MediatR;
+using Microsoft.Extensions.Logging;
+using SFA.DAS.Notifications.Messages.Commands;
+using SFA.DAS.Recruit.Application.Vacancies.Queries.GetVacancyById;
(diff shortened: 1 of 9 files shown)


src/Recruit/SFA.DAS.Recruit/Handlers/SharedApplicationReviewedEventHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+                    x.Name,
+                    vacancy.VacancyReference!.Value.ToString(),
+                    vacancy.EmployerName,
+                    string.Format(emailEnvironmentHelper.ManageVacancyProviderUrl, vacancy.TrainingProvider!.Ukprn!.Value, notification.VacancyId),
+                    string.Format(emailEnvironmentHelper.NotificationsSettingsProviderUrl, vacancy.TrainingProvider!.Ukprn!.Value)
+                ))
+                .Select(email => new SendEmailCommand(email.TemplateId, email.RecipientAddress, email.Tokens))
+                .Select(notificationService.Send).ToList();
+
+            if (emailTasks?.Count != 0)
```
killij, line 59: I think this should be emailTasks?.Count > 0, because otherwise null != 0.
This probably means that VacancySubmittedEventHandler also needs that updating too.

PR #2739, opened 2026-03-17
Csp 2417 update provider legal name in roatp v1

PeterDavidWilliams:


src/Roatp/SFA.DAS.Roatp/Application/Providers/Commands/UpdateProviderNamesCommandHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            var result = matchingProviderRecords.Select(matchingProvider => (ProviderAddress)matchingProvider)
+                .ToList();
+
+            var resultsFound = new GetUkrlpDataQueryResponse
+            {
+                Success = true,
+                Results = result
+            };
+            return await Task.FromResult(resultsFound);
```
DevenKShah: same here, no need to await

DevenKShah (dismissed):

src/Roatp/SFA.DAS.Roatp/Domain/Models/GetOrganisationsQueryResult.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Collections.Generic;
+
+namespace SFA.DAS.Roatp.Domain.Models;
+
+public class GetOrganisationsQueryResult
+{
+    public List<OrganisationModel> Organisations { get; set; }
```
raviudari01: Can this list initialized to empty list to avoid null issues ?

raviudari01 (dismissed):

raviudari01 (approved):

PR #2794, opened 2026-04-13
Flp 1743 payments integration issues

chrisfoster76:


src/LearnerData/SFA.DAS.LearnerData/Application/UpdateShortCourse/UpdateShortCourseLearningCommandHandler.cs
tom-gough, line 160: is this not in the 1739 branch?!

tom-gough (changes requested):

tom-gough (approved):

PR #2804, opened 2026-04-20
P2 2623 certificate matching

ratheeshri1984:


src/DigitalCertificates/SFA.DAS.DigitalCertificates/Application/Commands/CreateUserMatch/CreateUserMatchCommand.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public Guid UserId { get; set; }
+        public long? Uln { get; set; }
+        public string FamilyName { get; set; }
+        public DateTime DateOfBirth { get; set; }
+        public string CertificateType { get; set; }
+        public string CourseCode { get; set; }
+        public string CourseName { get; set; }
+        public string CourseLevel { get; set; }
+        public int? DateAwarded { get; set; }
```
ChrisJWoodcock, line 16: Why is this an int? - not a DateTime

src/DigitalCertificates/SFA.DAS.DigitalCertificates/Application/Queries/GetCertificatesMatch/GetCertificatesMatchQueryHandler.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            // Step 2: Search for certificate matches for each family name + date of birth
+            var allMatches = new List<CertificateMatchResult>();
+
+            if (identity.Identity != null && identity.DateOfBirth.HasValue)
+            {
+                var familyNames = identity.Identity
+                    .Select(i => i.FamilyName)
+                    .Where(n => !string.IsNullOrWhiteSpace(n))
+                    .Distinct()
```
ChrisJWoodcock: Should this be case insensitive are we matching on case sensitive names in assessor?

src/DigitalCertificates/SFA.DAS.DigitalCertificates.Api/Controllers/UsersController.cs
```diff
@@ -110,5 +135,52 @@ public async Task<IActionResult> CreateUserAction([FromRoute] Guid userId, [From
...
             }
         }
+
+        [HttpGet("{userId}/actions")]
+        public async Task<IActionResult> GetUserActions([FromRoute] Guid userId)
+        {
+            try
+            {
+                var result = await _mediator.Send(new GetUserActionsQuery { UserId = userId });
+                return Ok(new { useractions = result.UserActions });
```
ChrisJWoodcock, line 145: userActions?

src/DigitalCertificates/SFA.DAS.DigitalCertificates/Application/Queries/GetCertificatesMatch/GetCertificatesMatchQueryHandler.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                return new GetCertificatesMatchResult();
+            }
+
+            // Step 3: Collect all ULNs (new matches plus previously excluded) for mask retrieval
+
+            var standardUlns = allMatches
+                .Where(m => IsCertificateType(m.CertificateType, CertificateType.Standard))
+                .Select(m => m.Uln)
+                .Union(excludedUlns)
```
ChrisJWoodcock: By doing union here this is adding the excludedUlns to an empty list of standard matches and the same for framework so it will always get masks for each type even when there is no matches for that type if there some excludedUlns

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

src/DigitalCertificates/SFA.DAS.DigitalCertificates/Application/Commands/CreateUserMatch/CreateUserMatchCommand.cs
ChrisJWoodcock: It is only the year of DateAwarded being stored - field name will be updated to make that clear later

PR #2805, opened 2026-04-20
CSP 2446 outer is active available standard

afzal-hussain259:


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement/InnerApi/Responses/GetStandardResponse.cs
DevenKShah, line 8: This looks like unnecessary middleman.

afzal-hussain259: GetStandardResponse is required as its used in a few different places however i have refactored and removed GetStandardResponseFromCourseManagementApi  model and replaced with GetStandardResponse  as this looked unnecessary

DevenKShah (dismissed):

DevenKShah (dismissed):

src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.UnitTests/InnerApi/Responses/GetStandardResponseTests.cs
```diff
@@ -66,50 +66,257 @@ public void ImplicitConversion_FromCoursesApiWithFunding_MapsPropertiesAndUsesMo
...
+                EffectiveFrom = dateTimeNow.AddDays(-1),
+            }
+        };
+
+        // Act
+        GetStandardResponse result = source;
+
+        // Assert
+        result.IsActiveAvailable.Should().BeTrue();
     }
```
DevenKShah: All of these test methods are following the same lines of code, ideally these should be easily achieved using TestCase instead if individual methods.

DevenKShah (dismissed):

src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.UnitTests/InnerApi/Responses/GetStandardResponseTests.cs
afzal-hussain259: I tried putting these into test cases as below but test were not behaving as expected. When debugging the outcome would be correct but test fails with different outcome. Seems its issue with passing date formats. With the interest of time it maybe best to keep them separate for time being.

 [TestCase(null, null, false, TestName = "LastDateStartsIsNullAndEffectiveFromIsNull")]
 [TestCase(null, -1, true, TestName = "LastDateStartsIsNullAndEffectiveFromIsPastDate")]
 [TestCase(1, 0, true, TestName = "LastDateStartsIsFutureDateAndNotSameAsEffectiveFrom")]
 [TestCase(1, null, false, TestName = "LastDateStartsIsFutureDateAndEffectiveFromIsNull")]
 [TestCase(1, 1, false, TestName = "LastDateStartsIsFutureDateAndIsSameAsEffectiveFrom")]
 [TestCase(-1, 0, false, TestName = "LastDateStartsIsPastDate")]
 [TestCase(0, -1, true, TestName = "LastDateStartsIsToday")]
 [TestCase(1, 0, true, TestName = "EffectiveFromIsToday")]
 [TestCase(1, 2, false, TestName = "EffectiveFromIsFutureDate")]
 [TestCase(null, 1, false, TestName = "LastDateStartsIsNullEffectiveFromIsFutureDate")]
 public void ImplicitConversion_FromCoursesApi_IsActiveAvailable_TestCases(
     int? lastDateStartsNoOfDays,
     int? effectiveFromNoOfDays,
     bool expectedIsActiveAvailable)
 {
     // Arrange
     DateTime dateTimeNow = DateTime.UtcNow.Date;

     DateTime? lastDateStarts = lastDateStartsNoOfDays.HasValue ? dateTimeNow.AddDays(lastDateStartsNoOfDays.Value) : (DateTime?)null;
     DateTime? effectiveFrom = effectiveFromNoOfDays.HasValue ? dateTimeNow.AddDays(effectiveFromNoOfDays.Value) : (DateTime?)null;

     var source = new GetStandardResponseFromCoursesApi()
     {
         CourseDates = new CourseDates()
         {
             LastDateStarts = lastDateStarts,
             EffectiveFrom = effectiveFrom,
         }
     };

     // Act
     GetStandardResponse result = source;

     // Assert
     result.IsActiveAvailable.Should().Be(expectedIsActiveAvailable);
 }

DevenKShah: Its  a shame, we could have got this working.

DevenKShah (dismissed):

raviudari01 (approved):

PR #2823, opened 2026-04-27
Flp 1289 adjust english and maths

MikeV-TalentConsulting:


src/LearnerData/SFA.DAS.LearnerData/Services/UpdateLearningPutRequestBuilder.cs
```diff
@@ -152,4 +153,14 @@ private List<MathsAndEnglishDetails> BuildEnglishAndMathsDetails(List<MathsAndEn
...
     }
+
+    private static int? ResolveCombinedFundingAdjustmentPercentage(int? priorLearningAdjustment, int? otherFundingAdjustment)
+    {
+        if (!priorLearningAdjustment.HasValue && !otherFundingAdjustment.HasValue)
+        {
+            return null;
+        }
+
+        return priorLearningAdjustment.GetValueOrDefault(0) * otherFundingAdjustment.GetValueOrDefault(0);
```
chrisfoster76: Is it correct to default null values to zero when multiplying them?

Suppose we had a record where priorLearningAdjustment was null, and otherFundingAdjustment was 50%. We would combine them to make zero wouldn't we?

chrisfoster76 (changes requested):

chrisfoster76 (dismissed):

src/LearnerData/SFA.DAS.LearnerData/Requests/EarningsInner/UpdateEnglishAndMathsApiPutRequest.cs
```diff
@@ -26,6 +26,6 @@ public class EnglishAndMathsItem
     public DateTime? PauseDate { get; set; }
     public DateTime? WithdrawalDate { get; set; }
     public DateTime? CompletionDate { get; set; }
-    public int? CombinedFundingAdjustmentPercentage { get; set; }
+    public decimal? CombinedFundingAdjustmentPercentage { get; set; }
```
tom-gough, line 29: assume it came from the BAs that we should support non-integer percentages for these values?

tom-gough (dismissed):

tom-gough (approved):

PR #2840, opened 2026-05-05
[AWARD-1164] Blob metadata stored in DB

karen-hanna:


src/Aodp/SFA.DAS.Aodp/Application/Commands/Files/CreateFileMetadataCommandHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        public async Task<BaseMediatrResponse<EmptyResponse>> Handle(CreateFileMetadataCommand command, CancellationToken cancellationToken)
+        {
+            var response = new BaseMediatrResponse<EmptyResponse>() { Success = false };
+
+            try
+            {
+                var metadata = await _apiClient.PostWithResponseCode<EmptyResponse>(new CreateFileMetadataApiRequest() { Data = command });
+
+                response.Value = metadata.Body;
+                response.Success = true;
```
ChrisJWoodcock: This doesn't check for 4xx / 5xx error message so it assumes it worked

ChrisJWoodcock (changes requested):

src/Aodp/SFA.DAS.Aodp/Application/Commands/Files/CreateFileMetadataCommandHandler.cs
karen-hanna: added call to EnsureSuccessStatusCode

ChrisJWoodcock (approved):

PR #2847, opened 2026-05-07
CSP-2580-CSP-2582 ProviderPR and EmployerPR solution files updated for shared projects

raviudari01:
Removed SharedOuterApi projects and added SharedOuterApi.Types and its unit tests to both solution files. Updated Visual Studio version to 18. Fixed failing test GetRelationsByUkprnAornPayeTests


DevenKShah (commented):
Need to update read me with .net version. Also make sure other details are fine on read me.

DevenKShah (approved):

PR #2853, opened 2026-05-11
Appman 2281 add learners

witpo:


src/Approvals/SFA.DAS.Approvals/Application/Learners/Queries/GetLearnersForProviderQueryHandler.cs
```diff
@@ -81,8 +81,12 @@ public async Task<GetLearnersForProviderQueryResult> Handle(GetLearnersForProvid
             {
                 throw new ApplicationException($"Getting Draft Apprenticeships Failed. Status Code {draftApprenticeshipsResponse.StatusCode} Error : {draftApprenticeshipsResponse.ErrorContent}");
             }
-            var selectedUlns = draftApprenticeshipsResponse.Body.DraftApprenticeships.Select(x => x.Uln).ToList();
-            request.ExcludeUlns = selectedUlns;
+            var selectedUlns = draftApprenticeshipsResponse.Body.DraftApprenticeships
+                .Where(x => long.TryParse(x.Uln, out _))
+                .Select(x => long.Parse(x.Uln))
+                .ToList();
```
Pauljgraham, line 87: Just an observation, would it be better passing the Ulns as an integer array at the FE? Saving having to do this. But this is not a big concern

Pauljgraham (dismissed):
one tiny suggestion, but no worries if you choose to ignore it

src/Approvals/SFA.DAS.Approvals/Application/Learners/Queries/GetLearnersForProviderQueryHandler.cs
witpo: draft apprenticeship stores ULNs as string, I am using long in other locations that I have added

Pauljgraham (approved):

PR #2876, opened 2026-05-19
Flp 1801 change of provider

chrisfoster76:


src/LearnerData/SFA.DAS.LearnerData.Api.AcceptanceTests/Steps/RemoveShortCourseSteps.cs
```diff
@@ -0,0 +1,139 @@
...
+        var removedEpisodeKey = scenarioContext.Get<Guid>(EpisodeKey);
+        var requests = testContext.EarningsApi.MockServer.LogEntries;
+
+        requests.Should().ContainSingle(
+            r => r.RequestMessage.Url.Contains($"/{learningKey}/shortCourses/{removedEpisodeKey}") &&
+                 r.RequestMessage.Method == "DELETE",
+            "Expected a DELETE request to the Earnings domain for the short course");
+    }
+
+    private void ConfigureLearningApi()
```
tom-gough, line 54: Are these configure methods not re-usable outside the scope of this particular step file?

tom-gough (changes requested):

src/LearnerData/SFA.DAS.LearnerData.Api.AcceptanceTests/Steps/RemoveShortCourseSteps.cs
chrisfoster76: They configure the api explicitly for the removal scenario. In theory I suppose we might want to re-use them in future but could change the access modifier at that point.

tom-gough (dismissed):

tom-gough (approved):

PR #2892, opened 2026-05-27
updating to .Net 10

MS-DFE:


src/ApprenticeshipsManage/SFA.DAS.ApprenticeshipsManage.Api/SFA.DAS.ApprenticeshipsManage.Api.csproj
```diff
@@ -16,6 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Shared\SFA.DAS.SharedOuterApi.Types\SFA.DAS.SharedOuterApi.Types.csproj" />
```
cofaulco: don't need to reference in here as its pulled in transiently

cofaulco (changes requested):

cofaulco (dismissed):

cofaulco (approved):

PR #2897, opened 2026-05-28
P2-3236-Software Version Numbers Revealed

srinivasdudam-dfe:


ChrisJWoodcock:
@srinivasdudam-dfe Need to address the sonar cloud warning I am reading this as if you don't do that you are removing some default

ChrisJWoodcock (approved):

src/Assessors/SFA.DAS.Assessors.Api/web.config
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
```
ChrisJWoodcock, line 1: This is the wrong file - this should be a change to the assessor Inner API which is in the das-assessor-service repo.

ChrisJWoodcock (dismissed):

ChrisJWoodcock (approved):

PR #2917, opened 2026-06-04
APPMAN-2574-ILR-Stop-Event

Pauljgraham:


src/Approvals/SFA.DAS.Approvals.Api/Models/Apprentices/GetManageApprenticeshipDetailsResponse.cs
```diff
@@ -56,6 +50,7 @@ public class ApprenticeshipDetails
         public string ProviderReference { get; set; }
         public short Status { get; set; }
         public DateTime? StopDate { get; set; }
+        public int? WithdrawnReasonCode { get; set; }
```
cofaulco, line 53: No test asserts WithdrawnReasonCode propagates from inner GetApprenticeshipResponse to GetManageApprenticeshipDetailsResponse.Apprenticeship.WithdrawnReasonCode

cofaulco (changes requested):

src/Approvals/SFA.DAS.Approvals.Api/Models/Apprentices/GetManageApprenticeshipDetailsResponse.cs
Pauljgraham: I thought that, but the mapping is done via automapper and here is the link where the tests are preformed

https://github.com/SkillsFundingAgency/das-apim-endpoints/blob/276b0a735a26fb778a349ff18bc4deb59f9b82de/src/Approvals/SFA.DAS.Approvals.Api.UnitTests/Controllers/Apprentices/WhenGettingManageApprenticeshipDetails.cs?plain=1

cofaulco (approved):

PR #2940, opened 2026-06-15
FAI-3393 update recruit jobs outer api to use recruit contracts

killij:


balaji-j:
Not required as the other branch covers all the above changes.

PR #2948, opened 2026-06-17
P2-3365 Add family name and dob to match (Priority Fix)

ChrisJWoodcock:


ratheeshri1984 (approved):

ChrisJWoodcock:
This has been reverted as storing PII data in the Redis cache is not secure, replacing it with storing the user identity value

PR #2956, opened 2026-06-19
CSP-2624 Implement get restricted courses endpoint

afzal-hussain259:


src/RoatpCourseManagement/SFA.DAS.RoatpCourseManagement.UnitTests/Application/RestrictedCourses/Queries/GetAllRestrictedCourses/GetAllRestrictedCoursesQueryHandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        [Frozen] Mock<IRoatpCourseManagementApiClient<RoatpV2ApiConfiguration>> apiClientMock,
+        GetAllRestrictedCoursesQuery query,
+        GetAllRestrictedCoursesQueryHandler sut)
+    {
+        // Arrange
+        apiClientMock.Setup(c => c.GetWithResponseCode<GetAllRestrictedCoursesResponse>(It.IsAny<GetAllRestrictedCoursesRequest>())).ReturnsAsync(new ApiResponse<GetAllRestrictedCoursesResponse>(null, HttpStatusCode.InternalServerError, ""));
+
+        // Act & Assert
+        Assert.ThrowsAsync<ApiResponseException>(() => sut.Handle(query, new CancellationToken()));
+    }
```
PeterDavidWilliams, line 66: Do we want to throw on error or are we better off handling the error and returning (probably) empty?

afzal-hussain259: Hey Pete, Im using the EnsureSuccessStatusCode method to validate the status code. From inner we should only get 200 response so this check will validate any other errors like internal server errors.

PeterDavidWilliams (approved):

PR #2985, opened 2026-07-02
CSP-2627 Implemented outer api for get allowed providers by course

afzal-hussain259:


src/AdminRoatp/SFA.DAS.AdminRoatp.Api/Controllers/CourseController.cs
```diff
@@ -0,0 +1,22 @@
...
+using SFA.DAS.AdminRoatp.InnerApi.Responses;
+
+namespace SFA.DAS.AdminRoatp.Api.Controllers;
+
+[ApiController]
+[Route("/courses")]
+public class CourseController(IMediator _mediator, ILogger<CourseController> _logger) : ControllerBase
+{
+    [HttpGet("{larsCode}/providers/allowed")]
+    [ProducesResponseType(typeof(GetAllowedProvidersResponse), StatusCodes.Status200OK)]
```
raviudari01, line 13: 400 or 404 response type expected here ?

afzal-hussain259: This will return ok in outer only. There is a method in the handler called EnsureSuccessStatusCode which takes care of status codes

raviudari01 (approved):

PR #2993, opened 2026-07-07
P2 3484 feature admin consolidation

ratheeshri1984:


src/Admin/SFA.DAS.Admin/SFA.DAS.Admin.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MediatR" Version="14.1.0" />
+    <PackageReference Include="SFA.DAS.DigitalCertificates.Contracts" Version="1.0.9" />
```
ChrisJWoodcock: What package is this - the NuGet.Org ones are <1 with pre-release as they should be

ratheeshri1984: This is the one generated locally. After the review, I will create a NuGet package for the feature branch Inner API and reference it here.

ChrisJWoodcock (approved):

EllaMaggs (approved):

PR #3005, opened 2026-07-13
CSP-2002 use shared account users endpoint

raviudari01:


src/EmployerAan/azure-pipelines.yml
```diff
@@ -52,5 +52,7 @@ stages:
 - template: ../../pipeline-templates/stage/outerapi-pipeline.yml
   parameters:
     OuterApiName: EmployerAan
+    AdditionalProjectPathToInclude: src/Shared/SFA.DAS.SharedOuterApi.Employer.GovUK.Auth/SFA.DAS.SharedOuterApi.Employer.GovUK.Auth.csproj
+    AdditionalTestProjectPathToInclude: src/Shared/SFA.DAS.SharedOuterApi.Employer.GovUK.Auth.UnitTests/SFA.DAS.SharedOuterApi.Employer.GovUK.Auth.UnitTests.csproj
     ConfigurationSecrets:
```
DevenKShah, line 57: In all other solutions these projects have been directly referenced in to API project that way they get build along with it

DevenKShah (commented):
Package reference to Coverlet Collector is missing in unit test projects.

src/EmployerAan/azure-pipelines.yml
raviudari01: Here as well added in the solution but for pipeline build this is pattern 
image

DevenKShah (approved):

divesh987 (approved):

PR #3022, opened 2026-07-21
P2 3405 refactor

ratheeshri1984:


src/DigitalCertificates/SFA.DAS.DigitalCertificates.Api/Models/Users/GetUserResponse.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            return new GetUserResponse
+            {
+                Id = source.Id,
+                GovUkIdentifier = source.GovUkIdentifier,
+                EmailAddress = source.EmailAddress,
+                PhoneNumber = source.PhoneNumber,
+                LastLoginAt = source.LastLoginAt,
+                IsLocked = source.IsLocked,
+                DateOfBirth = source.DateOfBirth,
```
ChrisJWoodcock: The GetUserResponse in the Web does not need the DateOfBirth or Identities so no need to pass them back here - the spec will need updating

src/DigitalCertificates/SFA.DAS.DigitalCertificates.Api/Models/Users/UpdateUserIdentityRequest.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using SFA.DAS.DigitalCertificates.Application.Commands.UpdateUserIdentity;
+using SFA.DAS.DigitalCertificates.Models;
+
+namespace SFA.DAS.DigitalCertificates.Api.Models.Users
+{
+    public class UpdateUserIdentityRequest
```
ChrisJWoodcock, line 9: Not in your PR but can see that the GetUserIdentityResponse in APIM is using ValidFrom but the response in the inner API is using ValidSince (ValidSince is correct) this mismatch will cause an issue when there are multiple names so needs correcting

ChrisJWoodcock (changes requested):

ChrisJWoodcock (approved):

PR #3035, opened 2026-07-28
APPMAN-2545 : update employer shared UI package and outerapi types

skumardfe:


Pauljgraham (commented):
It's very odd that sharedUI package is in the APIM layer, what happens if we remove it?

Pauljgraham (approved):

